PR #16, opened 2021-03-09
make encrypted folder

haule24hdev:


haule24hdev:
> @haule24hdev this change will break mobiles apps, please return back data []byte

Yes, i am sorry. I will revert it now

PR #17, opened 2021-03-10
implement sync.go unit test

vudn95:
implement sync.go unit test


go.sum
```diff
@@ -1,98 +1,345 @@
...
+cloud.google.com/go v0.38.0/go.mod h1:990N+gfupTy94rShfmMCWGDn0LpTmnzTp2qbd1dvSRU=
+cloud.google.com/go v0.44.1/go.mod h1:iSa0KzasP4Uvy3f1mN/7PiObzGgflwredwwASm/v6AU=
+cloud.google.com/go v0.44.2/go.mod h1:60680Gw3Yr4ikxnPRS/oxxkBccT6SA1yMk63TGekxKY=
+cloud.google.com/go v0.45.1/go.mod h1:RpBamKRgapWJb87xiFSdk4g1CME7QZg3uwTez+TSTjc=
+cloud.google.com/go v0.46.3/go.mod h1:a6bKKbmY7er1mI7TEI4lsAkts/mkhTSZK8w33B4RAg0=
+cloud.google.com/go/bigquery v1.0.1/go.mod h1:i/xbL2UlR5RvWAURpBYZTtm/cXjCha9lbfbpx4poX+o=
+cloud.google.com/go/datastore v1.0.0/go.mod h1:LXYbyblFSglQ5pkeyhO+Qmw7ukd3C+pD7TKLgZqpHYE=
+cloud.google.com/go/firestore v1.1.0/go.mod h1:ulACoGHTpvq5r8rxGJ4ddJZBZqakUQqClKRT5SZwBmk=
+cloud.google.com/go/pubsub v1.0.1/go.mod h1:R0Gpsv3s54REJCy4fxDixWD93lHJMoZTyQ2kNxGRt3I=
+cloud.google.com/go/storage v1.0.0/go.mod h1:IhtSnM/ZTZV8YYJWCY8RULGVqBDmpoyjwiyrjsg+URw=
```
peterlimg, line 12: Are those cloud.google.com/go packages being used?

peterlimg (changes requested):
Clean up the go.sum to make sure that all packages are necessary.

go.sum
vudn95: Not sure, I just ran `go mod tidy`, then go sum changed. But I don't use that in my implementation. I will double-check those packages.

vudn95: Because I use `github.com/spf13/viper` to read config from yaml file. Let's have a see on `viper` package's go mod: https://github.com/spf13/viper/blob/master/go.sum

peterlimg: Alright. I find the viper is only used once in sync_test.go, I would recommend to use `gopkg.in/yaml.v2`, as it would only introduce 4 packages to go.sum instead of hundreds packages by viper.

peterlimg (changes requested):
The viper is only used once in sync_test.go, I would recommend to use gopkg.in/yaml.v2, as it would only introduce 4 packages to go.sum instead of hundreds packages by viper.

PR #22, opened 2021-03-23
Disallow uploads to allocations requiring > 32 blobbers

stewartie4:
As discussed in https://github.com/0chain/zboxcli/issues/14 the SDK implicitly supports max 32 blobbers per allocation due to the use of a 32-bit upload mask. 
Today, the SDK will either silently drop EC shards or never meet the required consensus when using these allocations.
Added validation to error out when an allocation requiring > 32 blobbers is used therefore preventing incomplete data persistence  

Validation:
20/30 EC combo: (previously would have dropped shards)
40/60 EC combo: (previously would have never met the required consensus)
image

10/20 EC combo:
image

### DIFF
--- a/zboxcore/sdk/allocation.go
+++ b/zboxcore/sdk/allocation.go
@@ -60,19 +60,19 @@ type BlobberAllocationStats struct {
 }
 
 type ConsolidatedFileMeta struct {
-	Name           	string
-	Type           	string
-	Path           	string
-	LookupHash     	string
-	Hash           	string
-	MimeType       	string
-	Size           	int64
-	ActualFileSize 	int64
+	Name            string
+	Type            string
+	Path            string
+	LookupHash      string
+	Hash            string
+	MimeType        string
+	Size            int64
+	ActualFileSize  int64
 	ActualNumBlocks int64
-	EncryptedKey   	string
-	CommitMetaTxns 	[]fileref.CommitMetaTxn
-	Collaborators  	[]fileref.Collaborator
-	Attributes     	fileref.Attributes
+	EncryptedKey    string
+	CommitMetaTxns  []fileref.CommitMetaTxn
+	Collaborators   []fileref.Collaborator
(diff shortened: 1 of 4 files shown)


zboxcore/sdk/uploadworker_test.go
```diff
@@ -0,0 +1,65 @@
+package sdk
```
stewartie4, line 1: I'm not sure if unit tests exist elsewhere (i couldn't find them) if they do let me know and i'll move these out

zboxcore/sdk/allocation.go
```diff
@@ -392,6 +395,10 @@ func (a *Allocation) uploadOrUpdateFile(localpath string, remotepath string,
 		uploadReq.fullconsensus = float32(bits.TrailingZeros32(uploadReq.uploadMask + 1))
 	}
 
+	if !uploadReq.IsFullConsensusSupported() {
+		return fmt.Errorf("allocation requires a greater number of blobbers than is supported. reduce number of data or parity shards and try again")
```
stewartie4, line 399: Requesting input on error message

sculptex: Looks good, however i'd suggest that the error message should also advise the actual limit.

zboxcore/sdk/uploadworker_test.go
stewartie4: 07f3f8837df504d2f5f164992b34f4542f60810e

zboxcore/sdk/allocation.go
stewartie4: 07f3f8837df504d2f5f164992b34f4542f60810e

stewartie4:
happy to merge? @sachin-0chain @sculptex @vudn95 @peterlimg

guruhubb:
Let me see if we can increase the limit. I'll assign this to @MurashovVen

PR #26, opened 2021-03-31
sign metadata and file transactions to blobber

stewartie4:
PR to support the changes to blobber security here:
https://github.com/0chain/blobber/pull/54/files
which in turn resolves https://github.com/0chain/blobber/issues/48


stewartie4:
closing due to increase in scope requiring an impl rethink

PR #27, opened 2021-04-02
Feature/increasing masks capacity

MurashovVen:
- Creating Uint128 type and writing unit tests for it.
- Changing upload and download masks type on Uint128.

PR #29, opened 2021-04-05
Revert "Disallow uploads to allocations requiring > 32 blobbers"

Sriep:
Reverts 0chain/gosdk#22


stewartie4:
What's the context on why this validation is being reverted? https://github.com/0chain/gosdk/pull/27 will not resolve the validation bug and I believe this needs to stay

Sriep:
Ok looks fine.

PR #32, opened 2021-04-23
add mcl-like map to G1

herumi:


mallochine:
alright I merged it into jssdk, let me try to test it real quick

PR #36, opened 2021-05-12
Setup test

vudn95:


Sriep:
Could you please provide a description of the purpose of this PR. 

It appears to add a pull request check, but none seems to be triggered in this PR. Also deletes a lot of tests, why? Adds some code to `allocation.go` and `http.go`; for what end? is there an attached issue?

PR #40, opened 2021-05-14
Return actual file size file info in RemoteFileList

iamrz1:
Upload files to dStorage using io reader (from local or remote locations). It helps to compare file sizes and make an informed decision.

### DIFF
--- a/zboxcore/sdk/sync.go
+++ b/zboxcore/sdk/sync.go
@@ -29,9 +29,10 @@ const (
 )
 
 type fileInfo struct {
-	Size int64  `json:"size"`
-	Hash string `json:"hash"`
-	Type string `json:"type"`
+	Size       int64  `json:"size"`
+	ActualSize int64  `json:"actual_size"`
+	Hash       string `json:"hash"`
+	Type       string `json:"type"`
 }
 
 type FileDiff struct {
@@ -51,7 +52,7 @@ func (a *Allocation) getRemoteFilesAndDirs(dirList []string, fMap map[string]fil
 			if _, ok := exclMap[child.Path]; ok {
 				continue
 			}
-			fMap[child.Path] = fileInfo{Size: child.Size, Hash: child.Hash, Type: child.Type}
+			fMap[child.Path] = fileInfo{Size: child.Size, ActualSize: child.ActualSize, Hash: child.Hash, Type: child.Type}
 			if child.Type == fileref.DIRECTORY {
 				childDirList = append(childDirList, child.Path)
 			}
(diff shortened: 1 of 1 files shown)


zboxcore/sdk/sync.go
```diff
@@ -29,9 +29,10 @@ const (
 )
 
 type fileInfo struct {
```
moldis, line 31: Actual Size is already added to fileRef
https://github.com/0chain/gosdk/commit/a7eeb212ef107e889f80bdf90c274b9c8f41019f

moldis (changes requested):
Please, check my comments, not sure if this branch needed

iamrz1:
Okay, that will do.

iamrz1:
I guess there are a few structs that returns fileInfo, this one is in sync.go. this struct is used in GetAllocation(). So I think we need this pr.

core/zcncrypto/bls0chain_test.go
```diff
@@ -58,7 +58,7 @@ func BenchmarkBLSSign(b *testing.B) {
 func TestRecoveryKeys(t *testing.T) {
 
 	sigScheme := NewSignatureScheme("bls0chain")
-    TestSignatureScheme(t)
+	TestSignatureScheme(t)
```
Sriep, line 61: No meaningful changes here, revert changes to this file.

zboxcore/sdk/allocation_test.go
```diff
@@ -31,7 +31,7 @@ func TestGetMinMaxWriteReadSuccess(t *testing.T) {
 		maxW, maxR, err := ssc.GetMaxWriteRead()
 		require.NoError(t, err)
 		require.Equal(t, 0.01, maxW)
-		require.Equal(t,0.01, maxR)
+		require.Equal(t, 0.01, maxR)
```
Sriep, line 34: No meaningful changes here, revert changes to this file.

core/zcncrypto/zcncrypto.go
```diff
@@ -46,6 +46,7 @@ type SignatureScheme interface {
 	// Combine signature for schemes BLS
 	Add(signature, msg string) (string, error)
 }
+
```
Sriep, line 49: No meaningful changes here, revert changes to this file.

Sriep (changes requested):
Three files with no real changes, need to be reverted. Only `sync.go` belongs in this PR.

Adding a field looks harmless, but I can't find anywhere it is used. Is there a link to it being used? Or is there a follow-up PR that will make use of the field?

iamrz1:
> Three files with no real changes, need to be reverted. Only `sync.go` belongs in this PR.
> 
> Adding a field looks harmless, but I can't find anywhere it is used. Is there a link to it being used? Or is there a follow-up PR that will make use of the field?

I removed the other fmt related changes.
There's a PR in zbox-cli (https://github.com/0chain/zboxcli/pull/21) that uses actual file size to compare existing file (in dStorage) with source file to determine if that upload was complete.

Sriep (approved):

PR #43, opened 2021-05-17
add setup file

vudn95:


zboxcore/sdk/common.go
```diff
@@ -6,15 +6,395 @@ import (
...
+)
+
+const (
+	configDir            = "test"
+	syncTestDir          = configDir + "/" + "sync"
+	syncDir              = syncTestDir + "/" + "sync_alloc"
+	textPlainContentType = "text/plain"
 )
 
+func blobberIDMask(idx int) string {
```
avanaur, line 32: If the changes added here on `common.go` are gonna be used for unit testing purposes only, perhaps put split it instead and add as `common_test.go`.

zboxcore/sdk/test/common/allocation.json
```diff
@@ -0,0 +1,124 @@
+{
```
avanaur, line 1: best to put the json files on a `testdata` folder designed to keep test files on go.

https://golang.org/pkg/cmd/go/internal/test/

```
The go tool will ignore a directory named "testdata", making it available
to hold ancillary data needed by the tests.
```

zboxcore/sdk/common.go
vudn95: Yes, I gree.

zboxcore/sdk/common_test.go
```diff
@@ -0,0 +1,393 @@
...
+	BlockWorker       string   `yaml:"block_worker"`
+	PreferredBlobbers []string `yaml:"preferred_blobbers"`
+	SignScheme        string   `yaml:"signature_scheme"`
+	ChainID           string   `yaml:"chain_id"`
+}
+
+func setupMockInitStorageSDK(t *testing.T, configDir string, blobberNums int) (miners, sharders []string, blobbers []*mocks.Blobber, close func()) {
+	var nodeConfig *nodeConfig
+
+	nodeConfigBytes := parseFileContent(t, configDir+"/"+"config.yaml", nil)
```
Sriep, line 69: There is no need to load config data from the file system. Just define a local nodeConfig object. somewhere.
Disk acess is expensive and messes up peoples hard drive with test data.

zboxcore/sdk/common_test.go
```diff
@@ -0,0 +1,393 @@
...
+	err := yaml.Unmarshal(nodeConfigBytes, &nodeConfig)
+	assert.NoErrorf(t, err, "Error yaml.Unmarshal(): %v", err)
+
+	clientBytes := parseFileContent(t, configDir+"/"+"wallet.json", nil)
+	clientConfig := string(clientBytes)
```
Sriep, line 74: There is no need to load clientConfig data from the file system. Just defie a lcoal clientConfig string. If it changes from test to test include it in `setupMockInitStorageSDK` input parameters somewhere.

zboxcore/sdk/common_test.go
```diff
@@ -0,0 +1,393 @@
...
+
+func writeFileContent(t *testing.T, fileName string, fileContentBytes []byte) {
+	fs, err := os.OpenFile(fileName, os.O_RDWR|os.O_CREATE, 0644)
+	assert.NoErrorf(t, err, "Error os.OpenFile() %v: %v", fileName, err)
+	defer fs.Close()
+	_, err = fs.Write(fileContentBytes)
+	assert.NoErrorf(t, err, "Error fs.Write() cannot write file content to %v: %v", fileName, err)
+}
+
+type nodeConfig struct {
```
Sriep, line 59: Why is this object needed. It only seems to be used to to define four objects, can't you set them directory. You could pass this to `setupMockInitStorageSDK` directly from the unit test if the values are going to change.

Sriep (changes requested):
We should be avoiding putting data on the file system. It's usually better to define objects in the tests. Access to the real file system should be mocked out wherever possible.
This PR is mainly about reading in data from the file system, which we should strive to avoid. I recommend looking for alternate solutions.

Sriep:
I am closing we should not be saving test data to disk in this way. Plus we can now mock httpClient.Go so the handler changes aren't needed.

PR #49, opened 2021-05-19
add filestatsworker.go test cases

vudn95:
Add test cases for filestatsworker.go

PR #48, opened 2021-05-18
add filemetaworker.go test cases

vudn95:
Add test cases for filemetaworker.go

PR #50, opened 2021-05-19
add listworker.go test cases

vudn95:
Add test cases for listworker.go


Sriep (changes requested):

zboxcore/sdk/listworker_test.go
```diff
@@ -0,0 +1,294 @@
...
+					Meta: map[string]interface{}{
+						"type": mockType,
+					},
+				},
+				blobberIdx:     41,
+				respStatusCode: http.StatusOK,
+			},
+			setup: func(t *testing.T, testName string, p parameters, _ string) {
+				mockClient.On("Do", mock.MatchedBy(func(req *http.Request) bool {
+					return req.URL.Path == "Test_Success"+zboxutil.LIST_ENDPOINT+mockAllocationTxId &&
```
Sriep, line 142: Both the "auth-token" and "path-hash" parameters are calculated by `getListInfoFromBlobber` so we should test that our function got it right.

Sriep (changes requested):

zboxcore/sdk/listworker_test.go
Sriep: you can access the `params` using 
```go
mock.MatchedBy(func(req *http.Request) bool {
   query := req.URL.Query()
   require.EqualValue(t, query.Get("auth_token"), whatever autho_toekn should be)
   require.EqualValue(t, query.Get("path_hash), whatever pah_hash shoul be)
   return req.URL.Path == "Test_Success"+zboxutil.LIST_ENDPOINT+mockAllocationTxId && 
...
```

PR #51, opened 2021-05-19
add renameworker.go test cases

vudn95:
Add test cases for renameworker.go


Sriep (changes requested):

Sriep (changes requested):

PR #54, opened 2021-05-24
add consensus.go test cases

vudn95:
Add test cases for consensus.go


Sriep (changes requested):

zboxcore/sdk/consensus_test.go
```diff
@@ -0,0 +1,166 @@
...
+		consensus       float32
+		consensusThresh float32
+		fullconsensus   float32
+	}
+	tests := []struct {
+		name   string
+		fields fields
+		want   bool
+	}{
+		// TODO: Add test cases.
```
Sriep, line 69: If more tests are needed add them.

zboxcore/sdk/consensus_test.go
```diff
@@ -0,0 +1,166 @@
...
```
Sriep, line 20: If more tests are needed add them.

Sriep (changes requested):
You added some TODOs is this still WIP? If not you can remove them.

PR #82, opened 2021-06-03
[FEATURE] Helper functions for eth wallets

moldis:
The branch contains helping functions to generate a wallet based on ETH wallet. It will be compiled to gomobile later on.


core/zcncrypto/ed255190chain.go
```diff
@@ -57,6 +57,11 @@ func (ed *ED255190chainScheme) GenerateKeys() (*Wallet, error) {
 	return w, nil
 }
 
+//GenerateKeysWithEth - not implemented
+func (ed *ED255190chainScheme) GenerateKeysWithEth(mnemonic, password string) (*Wallet, error) {
+	return nil, nil
```
Dmdv, line 62: I would return an error so not to confuse the client caller

core/zcncrypto/bls0chain.go
```diff
@@ -31,8 +31,26 @@ func NewBLS0ChainScheme() *BLS0ChainScheme {
 	return &BLS0ChainScheme{}
 }
 
+func (b0 *BLS0ChainScheme) GenerateKeysWithEth(mnemonic, password string) (*Wallet, error) {
```
Dmdv, line 34: Missing test for this new function

zcncore/ethwallet.go
```diff
@@ -0,0 +1,301 @@
+package zcncore
```
Dmdv, line 1: go test coverage 0.3% - could you pls add more tests?

Dmdv (changes requested):

zcncore/ethwallet.go
moldis: my latest commit is 67%

Sriep (approved):

PR #83, opened 2021-06-04
Add setup test

vudn95:


Sriep (changes requested):
You should not need this. Try doing the unit test without any of these changes. E.g. https://github.com/0chain/gosdk/pull/48

Sriep:
Closing as should be not be needed.

PR #119, opened 2021-06-21
Using grpc to communicate with the blobber

shravanshetty1:


kushthedude (changes requested):
Not ready to be integrated into master for now

Issue #120, opened 2021-06-22
Access add curator and curator transfer storagesc endponts


Sriep:
Access new endpoints for adding a curator and curator transfer allocation;  https://github.com/0chain/0chain/issues/369.

PR #121, opened 2021-06-22
Access to new curator endponts

Sriep:
Issue https://github.com/0chain/gosdk/issues/120, need to access new endpoints in https://github.com/0chain/0chain/pull/366

### DIFF
--- a/core/transaction/entity.go
+++ b/core/transaction/entity.go
@@ -99,6 +99,8 @@ const (
 	STORAGESC_UPDATE_ALLOCATION        = "update_allocation_request"
 	STORAGESC_WRITE_POOL_LOCK          = "write_pool_lock"
 	STORAGESC_WRITE_POOL_UNLOCK        = "write_pool_unlock"
+	STORAGESC_ADD_CURATOR              = "add_curator"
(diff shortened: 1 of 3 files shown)


core/transaction/entity.go
```diff
@@ -99,6 +99,8 @@ const (
 	STORAGESC_UPDATE_ALLOCATION        = "update_allocation_request"
 	STORAGESC_WRITE_POOL_LOCK          = "write_pool_lock"
 	STORAGESC_WRITE_POOL_UNLOCK        = "write_pool_unlock"
+	STORAGESC_ADD_CURATOR              = "add_curator"
+	STORAGESC_CURATOR_TRANSFER         = "curator_transfer_allocation"
```
andrenerd, line 103: imho, we could drop "curator" suffix here and in similar cases (see below). to make it:
```
STORAGESC_TRANSFER_ALLOCATION = "transfer_allocation"
```
why? because "curator" doesn't play to be reflected in naming. it's just a condition or invisible subject in the operation where "transfer" is action and "allocation" is object. in other cases we just reflect actions and objects - "add_curator", for example.

andrenerd (approved):

PR #126, opened 2021-06-25
[FEATURE] Create Dir with sdk method

moldis:
This method is related to 0Blobber PR for CreateDir.


moldis:
@lpoli I don't see any reason for status callback as this is fast operation. Status callback make sense for upload/download as there might be massive files size

lpoli:
@moldis 
A user issues a command "mkdir" in his terminal and 0fs will request this operation to the blobbers and upon successful response it should be displayed in the terminal/file_explorer, etc.

For zboxcli there is no much reason but for my case it would be good to have callback to all the commands so that for example; 0fs would wait "mkdir" command to complete. Otherwise I need to make possibly multiple request to fetch ObjectPath just to know the result.

Also status callback is minor change to make.

Sriep (approved):

PR #143, opened 2021-07-18
fix: remove redundant configs present in repo due to legacy

### LINKED ISSUE #142: chore: redundant editor configs added in the legacy
.idea & .vs-code are the editor configuration files created by the IDEs for local development, they are different for system to system on the basis of the configurations.

We should remove these files as they are redundant files, which serves no purpose in being added.

kushthedude:
fixes #142


kushthedude:
ping @Sriep

Sriep (approved):

PR #152, opened 2021-07-26
Integrating with pkg/errors

NoSkillGuy:
**Changes**:
1. using https://github.com/pkg/errors instead of our self implementation of errors!


kushthedude (commented):
Aren't you doing same thing here? Why to implement Wrap & Error struct when pkg/error has got everything. Just rip off the `core/common/errors` & use pkg/errors everywhere. Else there is no difference in this implementation & the older, its just redundant change to again give a wrapper over already implemented functionality.

NoSkillGuy:
@kushthedude we want to have our application error to have code and message.

NoSkillGuy:
@kushthedude this error package would wrap on `pkg/errors`. Here we can have some standard errors for the application.

NoSkillGuy:
Closing this PR in favour of https://github.com/0chain/gosdk/pull/196

PR #162, opened 2021-07-30
[BUG] Resolving bug with shared files

moldis:
Resolving bug with sharing files. When we trying to download shared files we sending AuthTicket - it contains field 'ContentHash', this 'ContentHash' picked up from `--share` function only from single blobber. 

Meantime blobber has check if 'ContentHash' != ref.ContentHash then it will reject request, that seems strange as ContentHash are not unique per blobbers and generated randomly (each blobber has it's own ContentHash for same file), altrought on blobber side this field seems useless for me.

Changed to ActualFileHash from both sides (gosdk and blobber repo) - this check happens only for Files, so it will not affect folder sharing

blobber PR: https://github.com/0chain/blobber/pull/278


guruhubb:
ContentHash is unique per blobber since the file is split up in to data and parity shards, and is not random because its the hash of the split up content. But you're right that we should use ActualFileHash which is the original content prior to being split up as a check between client (gosdk) and blobber to verify the content before serving it.

PR #183, opened 2021-08-05
137 grpc file rename

sajald77:
issue: [Integrating grpc equivalent of "/v1/file/rename/"](https://github.com/0chain/gosdk/issues/137)


kushthedude (commented):
I dont think the following tests have been run. Couldn't find any logs for your test in the raw test results https://pipelines.actions.githubusercontent.com/EqqyIn3qhFAeKJoAjnJkWFE8Y0C67kHL7wgWyaNRQowUMsANB4/_apis/pipelines/1/runs/711/signedlogcontent/4?urlExpires=2021-08-05T06%3A29%3A43.9056533Z&urlSigningMethod=HMACV1&urlSignature=kE2FOoA%2Bzx2IM73SlH6us3lUVf4qUOPvZZYDuSP0O98%3D

I think it will be better if we first fix the pipeline to make sure we dont commit any false positives.

core/clients/blobberClient/client.go
```diff
@@ -270,3 +271,27 @@ func CalculateHash(url string, req *blobbergrpc.CalculateHashRequest) ([]byte, e
...
+		blobbercommon.ClientHeader:          client.GetClientID(),
+		blobbercommon.ClientKeyHeader:       client.GetClientPublicKey(),
+		blobbercommon.ClientSignatureHeader: clientSignature,
+	}))
+
+	renameObjectResp, err := blobberClient.RenameObject(grpcCtx, req)
+	if err != nil {
+		return nil, err
+	}
+	return json.Marshal(renameObjectResp)
```
kushthedude, line 296: Dont we need to convert the object to the necessary type?

core/clients/blobberClient/client.go
```diff
@@ -270,3 +271,27 @@ func CalculateHash(url string, req *blobbergrpc.CalculateHashRequest) ([]byte, e
...
+
+	grpcCtx := metadata.NewOutgoingContext(context.Background(), metadata.New(map[string]string{
+		blobbercommon.ClientHeader:          client.GetClientID(),
+		blobbercommon.ClientKeyHeader:       client.GetClientPublicKey(),
+		blobbercommon.ClientSignatureHeader: clientSignature,
+	}))
+
+	renameObjectResp, err := blobberClient.RenameObject(grpcCtx, req)
+	if err != nil {
+		return nil, err
```
kushthedude, line 294: Can we follow error wrapping over here? Something like this
```suggestion
		return nil, errors.Wrap(err, "failed to RenameObject")
```

core/clients/blobberClient/client.go
```diff
@@ -270,3 +271,27 @@ func CalculateHash(url string, req *blobbergrpc.CalculateHashRequest) ([]byte, e
...
+
+func RenameObject(url string, req *blobbergrpc.RenameObjectRequest) ([]byte, error) {
+	blobberClient, err := newBlobberGRPCClient(url)
+	if err != nil {
+		return nil, err
+	}
+
+	clientSignature, err := client.Sign(encryption.Hash(req.Allocation))
+	if err != nil {
+		return nil, err
```
kushthedude, line 283: Same as below

kushthedude (commented):
Some minor changes, overall looks good!

core/clients/blobberClient/client.go
sajald77: There are two reasons for that:
- FIrst being that we do not have a response handler for renameObjectResp in "github.com/0chain/blobber/code/go/0chain.net/blobbercore/convert"
   We need to create one there first.
- Second being that we do not actually make any use of the response object as of right now.

So I'm not sure, if I have to create a response handler in the blobber repo, and then after it is updated, use in this one. What do you suggest? @kushthedude

core/clients/blobberClient/client.go
sajald77: we actually, handle that error outside the client function, in here:
`	_, err = blobberClient.RenameObject(blobber.Baseurl, &blobbergrpc.RenameObjectRequest{
		Path:         req.remotefilepath,
		Allocation:   req.allocationTx,
		ConnectionId: req.connectionID,
		NewName:      req.newName,
	})
	if err != nil {
		Logger.Error("could not rename object from blobber -" + blobber.Baseurl + " - " + err.Error())
		return nil, err
	}
`
I think this pattern is followed across other client functions as well. I think wrapping the error on a level above should be good as well.

core/clients/blobberClient/client.go
sajald77: @kushthedude I added the renameObjectResponse handler in blobber repo: https://github.com/0chain/blobber/pull/296
. I used that commit hash for the this commit for the blobber package, utilizing the newly created renameObjectResponse handler. Converting it to the necessary type.

kushthedude: Everything is already present in the grpc_v2 branch of blobber.

core/clients/blobberClient/client.go
kushthedude: Thats not wrapping the errors, that's just catching the errors. See this how wrapping works https://blog.golang.org/go1.13-errors , I have made sure that in blobber grpc handlers we wrap the errors whereever possible. Let's start doing this in the client too, we shouldn't repeat the mistakes whatever done in legacy

core/clients/blobberClient/client.go
sajald77: @kushthedude I couldn't find the responsehandler for rename object in grpc_v2 branch of the blobber.
It should be here right?
https://github.com/0chain/blobber/blob/grpc_v2/code/go/0chain.net/blobbercore/convert/response_handler.go

kushthedude (approved):
Can you update the branch to the latest grpc, this PR looks good! We can merge it after you have rebased the branch

kushthedude (commented):
Hey @sajald77, can you resolve the conflicts?

PR #193, opened 2021-08-06
Bug in error wrapping which is causing invalid wrap.

NoSkillGuy:


core/transaction/entity.go
```diff
@@ -224,7 +224,6 @@ func VerifyTransaction(txnHash string, sharders []string) (*Transaction, error)
 			continue
 		} else {
 			if response.StatusCode != 200 {
-				customError = errors.Wrap(customError, err)
```
cnlangzi, line 227: It is not good idea to skip error. The response error is hidden here. We don't know what happen on api request.

cnlangzi: It is better to refactor the logic here.

cnlangzi (changes requested):

core/transaction/entity.go
NoSkillGuy: Agreed, but that refactoring can go in a different PR. 

I've added this particular line in the previous commit, it wasn't there before.

cnlangzi:
It will be fixed on https://github.com/0chain/gosdk/issues/195. let us close it first.

PR #196, opened 2021-08-07
Integrating with 0chain errors package

NoSkillGuy:


go.mod
```diff
@@ -1,6 +1,7 @@
 module github.com/0chain/gosdk
 
 require (
+	github.com/0chain/errors v0.0.0-20210806215752-722ff90592d4 // indirect
```
cnlangzi, line 4: is it better to follow https://golang.org/doc/modules/version-numbers ? v0.0.0 means `In development`

cnlangzi (changes requested):

go.mod
NoSkillGuy: I directly took it from commit. will pull it from proper release tag . Thanks 👍🏼

NoSkillGuy: Now it's resolved

PR #201, opened 2021-08-10
Wrap 0chain remove-curator endpoint

Sriep:
Works with https://github.com/0chain/0chain/pull/470 needed for https://github.com/0chain/zboxcli/pull/83

### DIFF
--- a/core/transaction/entity.go
+++ b/core/transaction/entity.go
@@ -105,6 +105,7 @@ const (
 	STORAGESC_WRITE_POOL_LOCK          = "write_pool_lock"
 	STORAGESC_WRITE_POOL_UNLOCK        = "write_pool_unlock"
 	STORAGESC_ADD_CURATOR              = "add_curator"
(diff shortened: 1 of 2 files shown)


core/transaction/entity.go
```diff
@@ -105,6 +105,7 @@ const (
 	STORAGESC_WRITE_POOL_LOCK          = "write_pool_lock"
 	STORAGESC_WRITE_POOL_UNLOCK        = "write_pool_unlock"
 	STORAGESC_ADD_CURATOR              = "add_curator"
+	STORAGESC_REMOVE_CURATOR           = "remove_curator"
```
cnlangzi, line 108: Could we  use CamelCase instead of ALL_CAPS in Go names?

cnlangzi (changes requested):

core/transaction/entity.go
Sriep: Yes, but using the existing format takes preference.

PR #205, opened 2021-08-11
Revoke access for non encrypted shared files

moldis:
This commit will allow to use --revoke for non-encrypted files, same as for encrypted


cnlangzi (approved):
Please add comment as many as possible on coding.

PR #217, opened 2021-08-22
Initialize conf.cfg inside sdk.InitStorageSDK if it is possible

cnlangzi:


zboxcore/sdk/sdk.go
```diff
@@ -76,17 +76,50 @@ func GetLogger() *logger.Logger {
 	return &Logger
 }
 
-// InitStorageSDK init storage sdk with wallet json and config
-func InitStorageSDK(walletJSON string, cfg conf.Config) error {
-	err := client.PopulateClient(walletJSON, cfg.SignatureScheme)
+// InitStorageSDK init storage sdk with clientJSON
```
moldis, line 79: @cnlangzi I have been wrong here. InitStorageSDK we always passing walletJson and also in mobile Apps we calling this `Zcncore.init` where we actually passing chain config

cnlangzi: moved it to zcncore.Init instead of sdk.InitStorageSDK.

zcncore/wallet.go
```diff
@@ -340,12 +361,23 @@ func WithConfirmationChainLength(m int) func(c *ChainConfig) error {
 }
 
 // InitZCNSDK initializes the SDK with miner, sharder and signature scheme provided.
-func InitZCNSDK(blockWorker string, signscheme string, configs ...func(*ChainConfig) error) error {
-	if signscheme != "ed25519" && signscheme != "bls0chain" {
+func InitZCNSDK(cfg *conf.Config) error {
```
moldis, line 364: @cnlangzi I think InitZCNSDK is used in wasm 0proxy and it's very unlikely to use pointers and data structures in function params, very often they can't be converted to mobiles/wasm library. Consider leaving the same params as before and construct conf.Config inside function

zcncore/wallet.go
```diff
@@ -289,15 +290,35 @@ func CloseLog() {
...
+	err := json.Unmarshal([]byte(chainConfigJSON), &_config.chain)
 	if err == nil {
 		// Check signature scheme is supported
 		if _config.chain.SignatureScheme != "ed25519" && _config.chain.SignatureScheme != "bls0chain" {
 			return errors.New("", "invalid/unsupported signature scheme")
 		}
 
-		err := UpdateNetworkDetails()
+		// try to initiazlie conf/Config
+		reader, err := conf.NewReaderFromJSON(chainConfigJSON)
```
moldis, line 313: This code logic looks duplicated with code logic above to unmarshal config
`err := json.Unmarshal([]byte(chainConfigJSON), &_config.chain)`

core/conf/reader_test.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestJSONReader(t *testing.T) {
+
+	reader, _ := NewReaderFromJSON(`{
+		"chain_id":"0afc093ffb509f059c55478bc1a60351cef7b4e9c008a53a6cc8241ca8617dfe",
```
moldis, line 12: Could you please make values more dummy and not keep actual in the repo, even it's dev chain but it might give security issues

zcncore/wallet.go
Rizary: +1 to the comment. can we also pass the config in json or string instead of pointer?

zcncore/wallet.go
Rizary: can we just move the checking on `err := json.Unmarshal([]byte(chainConfigJSON), &_config.chain)` along with signature checking into `conf.NewReaderFromJSON(chainConfigJSON)` instead?

zboxcore/sdk/sdk.go
```diff
@@ -76,17 +75,29 @@ func GetLogger() *logger.Logger {
...
+	// try to initialize client for zbox
+	err := client.PopulateClient(walletJSON, signatureScheme)
 	if err != nil {
 		return err
 	}
-	blockchain.SetChainID(cfg.ChainID)
-	blockchain.SetPreferredBlobbers(cfg.PreferredBlobbers)
-	blockchain.SetBlockWorker(cfg.BlockWorker)
 
-	transaction.SetConfig(&cfg)
```
Rizary, line 89: when we remove this, where do we set the config for the transaction?

PR #225, opened 2021-09-01
added formatting helpers for balance (zcn,mzcn,uzcn,sas) #10

bbist:
This PR includes formatting helpers used to format balance (ZCN) to string for CLI applications. It addresses the issue raised here: https://github.com/0chain/zwalletcli/issues/10


zcncore/zcnfmt.go
```diff
@@ -0,0 +1,121 @@
...
+func (token Balance) Format(unit Unit) string {
+	switch unit {
+	case SAS:
+		return fmt.Sprintf("%d SAS", token)
+	case UZCN:
+		return fmt.Sprintf("%.3f uZCN", float64(token)/1e3)
+	case MZCN:
+		return fmt.Sprintf("%.3f mZCN", float64(token)/1e6)
+	}
+	return fmt.Sprintf("%.3f ZCN", float64(token)/1e9)
```
NoSkillGuy, line 75: According to this issue [here](https://github.com/0chain/zwalletcli/issues/10#issue-891467717)
```
0.0000000001 ZCN = 0.0000001 mZCN = 0.0001 µZCN = 1 sas 
```
reverse of this 
```
1 µZCN = 1e4 sas
1 mZCN = 1e7 sas
1 ZCN = 1e10 sas
```
But i see you have considered only `1e9`

sculptex: Indeed the testFormat() function is wrong. I think the confusion lies is Saswatas not being a 3rd power of decimals like usual thousand separators. It should use 1e4, 1e7 and 1e10. Saswatas shouldn't really be displayed if we set threshold for displaying uzcn to 4 decimals, e.g. 1 sas = 0.0001 µZCN

sculptex: Just to clarify, the format chosen by autoformat is a separate layer to the amount of raw zcn being shown.

That is, there could be user preference to show amounts as zcn down to 0.01 ZCN, then below that e.g. 0.0099 would be shown as 9.9mZCN.

bbist: @NoSkillGuy, thanks for pointing it out. I"ll fix the order as 1 zcn = 1e10 sas.

bbist:
> We could also do with a slightly different way to show these smaller denominations so they are more clearly distinct from ZCN in GUI environments.
> 
> I thought a different font colour and/or progressively smaller font size for smaller denominations might help. Could these formats be in some sort of helper function.
> 
> E.g. these are based on output denomination
> 
> getCol(zcn) returns default
> getCol(mzcn) returns orange
> getCol(uzcn) returns color blue
> etc.
> 
> getSize (zcn) returns 1 (100%)
> getSize (mzcn) returns 0.9 (90%)
> getSize(uzcn) returns 0.8 (80%)
> etc.

@sculptex, I guess I could add additional functions to format them into html as well as for CLI with color support. For a client side UI code, it makes sense to implement the same logic separately.

sculptex (changes requested):
Looks numerically correct now for values but only on mobile so can't check fully.

Since sas are lowest denomination, it doesn't make sense to have decimal places on those?
There is also scope on ZCN if value is large, e.g. 1,000,000ZCN you really aren't bothered with showing 3 decimal places. There could either be a straight cut off above a certain value, or decrease decimal places,
e.g.
<1,000 3 decimals
>=1,000 2 decimals
>=10,000 1 decimal
>=100,000 no decimals

bbist:
> Looks numerically correct now for values but only on mobile so can't check fully.
> 
> Since sas are lowest denomination, it doesn't make sense to have decimal places on those?
> There is also scope on ZCN if value is large, e.g. 1,000,000ZCN you really aren't bothered with showing 3 decimal places. There could either be a straight cut off above a certain value, or decrease decimal places,
> e.g.
> <1,000 3 decimals
> 
> > =1,000 2 decimals
> > =10,000 1 decimal
> > =100,000 no decimals

Removed decimal formatting from SAS, but left the 3 digit formatting on ZCN (unit) for now. I'll merge this PR and tag with a new version so that we can update the dependency on zwalletcli and zboxcli.

PR #232, opened 2021-09-05
Refactor BlockDownloadRequest

Sriep:
Refactor BlockDownloadRequest. Clean up the code a little, with the extra benefit that blobber download unit test can reuse the code to setup forms.


Sriep:
Would be nice but not necssary

PR #251, opened 2021-09-15
fix(zcncrypto):reverted bls to bls-go-binary

cnlangzi:
see detail on https://0chain.slack.com/archives/G01GMABG2T1/p1631608351009600


moldis (approved):
Looks good, but pls test with Ubuntu as I understood that's main reason of changing it

cnlangzi:
> Looks good, but pls test with Ubuntu as I understood that's main reason of changing it

OK. I will test it on Ubuntu

PR #252, opened 2021-09-16
Adding static methods for misc formatting

moldis:
Adding few extra methods as mobile/wasm clients can work only with primitive types


cnlangzi (approved):
LGTM

PR #263, opened 2021-10-07
Hotfix Temp path for all methods

moldis:
We need explicitly set temp path on clients. Consider to move it to config, as for now it's in methods


cnlangzi (approved):

cnlangzi (approved):
BTW, it is named as `workdir` on CreateChunkedUpload.

PR #267, opened 2021-10-19
re-implemented `CreateDir` and added development client toolkit

cnlangzi:
- Added new development client toolkit. see [detail](https://github.com/0chain/gosdk/blob/676f424e3ba61d9f33a61de36c48bc68dab10344/sdks/blobber/blobber.go#L34). It helps us to focus on request and response body without taking too much time on http 
- Re-implemented `CreateDir` feature on sdks
- Forwarded old `CreateDir`

linked to https://github.com/0chain/blobber/pull/395


zboxcore/sdk/dir_create.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"context"
+
+	"github.com/0chain/errors"
+	"github.com/0chain/gosdk/constants"
+	"github.com/0chain/gosdk/sdks/blobber"
+)
+
+// CreateDir create intermediate directories on blobbers
+func CreateDir(ctx context.Context, allocationID string, name string) error {
```
lpoli, line 12: I suppose this should a method of an allocation object just like other functions like; copy, move, list, etc

zboxcore/sdk/allocation.go
```diff
@@ -257,25 +268,9 @@ func (a *Allocation) UploadFile(localpath string, remotepath string,
 	return a.startChunkedUpload(localpath, remotepath, status, false, "", false, attrs)
 }
 
+// CreateDir [Deprecated]please use sdks.CreateDir
```
lpoli, line 271: why deprecated?

zboxcore/sdk/dir_create.go
cnlangzi: hmm, good question. it should be a personal favor.

- I like to make consumer easy to use sdk method. In zboxcli, it is easy to pass inputs to sdks.CreateDir without too much other jobs.
```
// use dir-create instead
var createDirCmd = &cobra.Command{
	Use:        "createdir",
	Deprecated: "please use mkdir",
	Short:      "Create directory in allocation",
	Long:       `Create directory in allocation`,
	Args:       cobra.MinimumNArgs(0),
	Run: func(cmd *cobra.Command, args []string) {
		fflags := cmd.Flags()              // fflags is a *flag.FlagSet
		if !fflags.Changed("allocation") { // check if the flag "path" is set
			PrintError("Error: allocation flag is missing") // If not, we'll let the user know
			os.Exit(1)                                      // and return
		}
		if !fflags.Changed("dirname") {
			PrintError("Error: dirname flag is missing")
			os.Exit(1)
		}

		allocationID := cmd.Flag("allocation").Value.String()
		dirname := cmd.Flag("dirname").Value.String()

		err := sdk.CreateDir(context.TODO(), allocationID, dirname)

		if err != nil {
			PrintError("ERR: ", err.Error())
			return
		}

		PrintInfof("OK: created directory '%s'", dirNewName)
	},
}

var dirNewCmd = &cobra.Command{
	Use:   "dir-new",
	Short: "Create directories named on blobbers",
	Long:  `Create directories named on blobbers`,
	Run: func(cmd *cobra.Command, args []string) {
		err := sdk.CreateDir(context.TODO(), dirNewAllocationID, dirNewName)

		if err != nil {
			PrintError("ERR: ", err.Error())
			return
		}

		PrintInfof("OK: created directory '%s'", dirNewName)
	},
}
  ```
- I don't like too much OOP in golang. Because golang is not design for OOP. I am trying refactor whole sdks one by one in `sdks` package.

File: zboxcore/sdk/allocation.go (hunk 2 above)
Comment (cnlangzi) [reply zboxcore/sdk/allocation.go:271]: because it is forwarded.

File: sdks/blobber/blobber.go [hunk 3]
```diff
@@ -2,20 +2,101 @@
...
-		ZBox:    zbox,
+		BaseURLs: baseURLs,
+		ZBox:     zbox,
 	}
 
 	return b
 }
+
+// CreateDir create new directories on blobbers
+func (b *Blobber) CreateDir(allocationID, name string) error {
```
moldis, line 34: @cnlangzi I think it should remain in the allocation object, just like other operations with files. More more we don't have access on clients to blobber object, so moving it here will broke all current implementation

sdks/zbox.go
```diff
@@ -79,3 +83,59 @@ func (z *ZBox) SignRequest(req *http.Request, allocationID string) error {
...
+
+	list := make([]string, len(baseURLs))
+	for k, v := range baseURLs {
+		list[k] = v + requestURL
+	}
+
+	return list
+}
+
+func (z *ZBox) DoPost(req *Request, handle resty.Handle) *resty.Resty {
```
moldis, line 122: zboxsdk already has lot's of methods for POST's why does it reimplemented here?

sdks/blobber/blobber.go
moldis: + clients have no idea what blobber is, all of them operating with allocation objects

zboxcore/sdk/dir_create.go
moldis: @cnlangzi I prefer to keep in allocation object. Bcz clients have no idea about blobber objects

zboxcore/fileref/fileref.go
```diff
@@ -114,33 +114,35 @@ func GetReferenceLookup(allocationID string, path string) string {
 
 func (r *Ref) CalculateHash() string {
 	if len(r.Children) == 0 && !r.childrenLoaded {
-		return r.Hash
+		// skip empty directory in hash
+		return ""
+		//return r.Hash
```
moldis, line 119: That's risky. Please test your approach with share directory function.

sdks/zbox.go
moldis: zboxutil.HttpDo

sdks/blobber/vars.go
```diff
@@ -0,0 +1,6 @@
+package blobber
+
+const (
+	// EndpointDirNew endpoint for create dire
+	EndpointDirNew = "/v1/dir/%s"
```
moldis, line 5: Why was this move alone from all other related methods with files?

sdks/blobber/blobber.go
```diff
@@ -2,20 +2,101 @@
...
+
+		return constants.ErrUnknown
+
+	})
+
+	urls := b.BuildUrls(b.BaseURLs, nil, EndpointDirNew, allocationID)
+
+	r.DoPut(context.TODO(), req.Body, urls...)
+
+	errs := r.Wait()
```
moldis, line 95: This call mostly will block the main thread on clients and cause ANR, previously we had a WaitGroup to avoid such issues.

moldis (changes requested):
Need some clarification

zboxcore/sdk/dir_create.go
lpoli: There is one issue with your approach. You are thinking it in terms of command operations. 0fs and 0s3-server uses allocation object to do most operations. 
These two projects makes a single request for getting an allocation object and using that object it will do other file operations. 
With your approach, for each file operation, a request for an allocation object is required. 

Also this kind of changes will affect all the repo. So `gosdk` and `blobber` changes should be made only after discussion with other members.

Optimization is not a problem, its always good to have. But for structural changes, many things have to considered.

sdks/blobber/blobber.go
lpoli: It is breaking 0s3, 0s3-server, 0fs, 0ipfs, etc.

core/resty/vars.go
```diff
@@ -6,7 +6,7 @@ var (
 	// DefaultDialTimeout default timeout of a dialer
 	DefaultDialTimeout = 5 * time.Second
 	// DefaultRequestTimeout default time out of a http request
-	DefaultRequestTimeout = 10 * time.Second
+	DefaultRequestTimeout = 0 * time.Second
```
lpoli: request timeout set to 0?

dev/blobber/init.go
```diff
@@ -10,4 +10,6 @@ func RegisterHandlers(s *mux.Router) {
 	s.HandleFunc("/v1/file/upload/{allocation}", uploadAndUpdateFile).Methods(http.MethodPut, http.MethodPost)
 	s.HandleFunc("/v1/file/referencepath/{allocation}", getReference).Methods(http.MethodGet)
 	s.HandleFunc("/v1/connection/commit/{allocation}", commitWrite).Methods(http.MethodPost)
+
+	s.HandleFunc("/v1/dir/{allocation}", createDir).Methods(http.MethodPut)
 }
```
lpoli, line 15: Is this `dev` directory for development purpose only?

zboxcore/sdk/dir_create.go
Sriep: I prefer to keep CreateDir in the allocation object, changes to existing interface objects should be kept to a minimum.

guruhubb: I agree. We're sprinting towards mainnet and should minimize disruption and optimization at this late stage. We have other higher priority issues.

PR #269, opened 2021-10-25
Fix/challenge logs

cnlangzi:
- fixed nested context cancel bug on resty
- simplied async http do on restry
- fixed reload bug on CompactMerkleTree


kushthedude (approved):
This looks good & more decoupled actually

PR #273, opened 2021-10-29
fix: rectify file stats info to have proper key

littlestar642:
Overview

While writing the file stats tests, I found that the stats returned were having a flaky nature in the sense the blobberID in the key was not always the same as the blobberID in the value. 

This PR is an attempt to rectify that issue. 

Solution

Previously the blobberIDs were getting filled as the key to the filestats map from the blobberID array. Now the array ascertained the sequence but the goroutines did not and there is always a race condition. As a result of this wrong blobberID was getting assigned to the stats for a different blobber. 

I am adding the blobberID from the stats itself, which eliminates the issue of wrongly assigned IDs.


stewartie4:
@littlestar642 / @Sriep / @Dmdv / @cnlangzi  does this look good to merge?

cnlangzi (approved):
LGTM

PR #274, opened 2021-10-30
Update allocaiont, dont set ower to client

Sriep:
Update allocation does not need to be sent by the allocation's owner as of PR620 https://github.com/0chain/0chain/pull/620 so stop GoSDK setting owner to the client.


Sriep:
Moved into https://github.com/0chain/gosdk/pull/288

PR #277, opened 2021-11-01
:bug: Unable to download encrypted file

moldis:
The issue happens if I want to download an encrypted file from my own storage - it was using logic from shared download instead. And shared download has a different encryption algorithm.

This PR should be merged along with the blobber PR: https://github.com/0chain/blobber/pull/409/files

### DIFF
--- a/zboxcore/sdk/blockdownloadworker.go
+++ b/zboxcore/sdk/blockdownloadworker.go
@@ -80,9 +80,6 @@ func InitBlockDownloader(blobbers []*blockchain.StorageNode) {
 	initDownloadMutex.Lock()
 	defer initDownloadMutex.Unlock()
 	if downloadBlockChan == nil {
-		// for _, v := range downloadBlockChan {
-		// 	close(v)
-		// }
 		downloadBlockChan = make(map[string]chan *BlockDownloadRequest)
 	}
 	blobberReadCounter = &sync.Map{}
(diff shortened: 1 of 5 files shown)


zboxcore/sdk/blockdownloadworker.go
```diff
@@ -192,26 +189,29 @@ func (req *BlockDownloadRequest) downloadBlobberBlock() {
...
 					rspData.Success = true
-					//rawData := make([]byte,0)
-					//json.Unmarshal(response, &rawData)
 					rspData.RawData = response
+
+					// TODO download by 'chunks' should not download full stream and then chunk it
+					// It has to download already chunked data
 					if len(req.encryptedKey) > 0 {
-						// 256 for the additional header bytes,  where chunk_size - 2 * 1024 is the encrypted data size
-						chunks := req.splitData(rspData.RawData, req.chunkSize-2*1024+256)
```
cnlangzi, line 214: why it is 256? encryption header size is  2*1024 + a 16 gap?

cnlangzi (changes requested):

zboxcore/sdk/blockdownloadworker.go
moldis: @cnlangzi Yes, I am surprised myself too. But this is legacy code that works with auth_tickets

cnlangzi: ok. ignore it first.

cnlangzi (approved):

PR #283, opened 2021-11-02
Feature/zcnsc

Dmdv:


constants/files.go
```diff
@@ -2,7 +2,7 @@ package constants
 
 const (
 	// FileOperationInsert method name of add new file
-	FileOperationInsert      = "insert"
+	FileOperationInsert      =x "insert"
```
moldis: @Dmdv Seems error here

.golangci.yml
```diff
@@ -0,0 +1,79 @@
+linters-settings:
```
moldis: @Dmdv I think this file should be in different branch

zcnbridge/authorizers_query.go
```diff
@@ -0,0 +1,213 @@
...
+
+	quorum := math.Ceil((float64(numSuccess) * 100) / float64(totalWorkers))
+
+	if numSuccess > 0 && quorum >= wallet.ConsensusThresh && len(results) > 1 {
+		burnTicket, ok := results[0].Data().(*proofEthereumBurn)
+		if !ok {
+			return nil, errors.Wrap("type_cast", "failed to convert to *proofEthereumBurn", err)
+		}
+
+		var sigs []*zcnsc.AuthorizerSignature
```
moldis, line 150: As I understood Authorizers should authorize burn tickets, how long it will take if the client calls it? And how many Authorizers required?

zcnbridge/bridge.go
```diff
@@ -0,0 +1,334 @@
...
+
+	// To
+	tokenAddress := common.HexToAddress(config.Bridge.WzcnAddress)
+
+	gasLimit, err := etherClient.EstimateGas(context.Background(), eth.CallMsg{
+		To:   &tokenAddress, // FIXME: From: is required?
+		Data: data,
+	})
+	if err != nil {
+		zcncore.Logger.Fatal(err)
```
moldis: Wrong return value

constants/files.go
Dmdv: Fixed

.golangci.yml
Dmdv: Fixed

zcnbridge/authorizers_query.go
Dmdv: It depends on the transaction confirmed. If Ethereum transaction is confirmed then this will be fast 1-10 seconds. For ethereum transaction this is fast. We will have for start 20 authorizers. All authorizer requests will be run async concurrently @moldis

zcnbridge/bridge.go
Dmdv: Fixed

zcnbridge/authorizers_query.go
moldis: @Dmdv Even 10 seconds is slow, and I guess it might take more time and not depend on us, more on eth network. Any way we can move this to the backend side and let client just sign the transaction, then give back some sort of hash to check transactions status?

PR #286, opened 2021-11-04
Fix/sync upload for 0proxy

cnlangzi:
- updated sdk to support to work on 0Proxy

### DIFF
--- a/zboxcore/sdk/live_upload_reader_youtubedl.go
+++ b/zboxcore/sdk/live_upload_reader_youtubedl.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


cnlangzi:
fixed https://github.com/0chain/roadmap/issues/44

zboxcore/sdk/live_upload_reader_youtubedl.go
```diff
@@ -102,15 +101,15 @@ func (r *YoutubeDL) wait() {
 func (r *YoutubeDL) Close() error {
 	if r != nil {
 		if r.cmd != nil {
-			r.cmd.Process.Kill()
+			r.cmd.Process.Kill() //nolint
```
kushthedude, line 104: What does the lint complain over here?

kushthedude (approved):
looks good to me, just a small query

zboxcore/sdk/live_upload_reader_youtubedl.go
cnlangzi: because there is an `error` output. but we don't need to check it there.

PR #290, opened 2021-11-08
Use PublicKey for VerifySignature function

rizary:
~~This PR should be merged after  we make sure all blobber test regarding Miracl signature and conversion are passed.~~

It's hard to test blobber without this PR getting merged. What do you think @cnlangzi ?


cnlangzi (approved):
LGTM

PR #292, opened 2021-11-09
Fix/improved js bridge on wasm

cnlangzi:
- added jsbridge package to speed up wasm development


wasmsdk/jsbridge/func_test.go
```diff
@@ -0,0 +1,87 @@
+// +build js,wasm
```
rizary: Last time I check, adding test file with `+build js,wasm` make the output file bigger. Please double check it again. If it is the case, then I propose not include the test file to the final build. If not, then ignore this comment.

cnlangzi: tested it. they seems be same size. because test files are not built on `go build`. they work with `go test` only.

rizary (approved):
LGTM.

PR #293, opened 2021-11-10
fix(zcncrypto):rollback to hermui bls

cnlangzi:
rollback to hermui bls first. it blocks everything. It will be replaced with miracl bls after it is test passed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ wasm-test:
 gosdk-mocks:
 	./generate_mocks.sh
 
-gosdk-test: wasm-test
+gosdk-test:
 	go test -v -tags bn256 $(shell go list ./... | grep -v wasm)
 
 install-gosdk: | gosdk-build wasm-build
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ gosdk-mocks:
 	./generate_mocks.sh
 
 gosdk-test:
-	go test -v -tags bn256 $(shell go list ./... | grep -v wasm)
+	go test -v -tags bn256 ./...
 
 install-gosdk: | gosdk-build wasm-build
 
--- a/core/zcncrypto/bls0chain_test.go
+++ b/core/zcncrypto/bls0chain_test.go
@@ -1,157 +1,19 @@
 package zcncrypto
 
 import (
-	"bytes"
-	"fmt"
 	"testing"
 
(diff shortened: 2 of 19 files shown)


core/zcncrypto/bls0chain_test.go
```diff
@@ -219,108 +72,82 @@ func TestRecoveryKeys(t *testing.T) {
...
-// 	pk1 := MiraclToHerumiPK(miraclpk1)
-//
-// 	require.EqualValues(t, pk1, "68d37ed84842c91d9f82389489a1b1a7ab7a957816c635ee750769853aeddf1b490b3aa185a3f01f537cd1e9e56c43ab2617c8a3f9d2a1fd0dae23d26b2ca018")
-//
-// 	// Assert DeserializeHexStr works on the output of MiraclToHerumiPK
-// 	var pk bls.PublicKey
-// 	err := pk.DeserializeHexStr(pk1)
-// 	require.NoError(t, err)
-// }
+func TestMiraclToHerumiPK(t *testing.T) {
```
rizary, line 143: I think `TestMiraclToHerumiPK` is not used outside gosdk? I remember blobber has its own `TestMiraclToHerumiPK`. It's fine to commented it.

rizary:
You should rename `Mircal` into `Miracl` to avoid confusion in the future. I see couples of code name with mircal.

Makefile
```diff
@@ -35,8 +35,8 @@ wasm-test:
 gosdk-mocks:
 	./generate_mocks.sh
 
-gosdk-test: wasm-test
-	go test -v -tags bn256 $(shell go list ./... | grep -v wasm)
+gosdk-test:
```
rizary, line 38: did we remove the wasm test in ci?

rizary (approved):
LGTM if all the above changes is done.

core/zcncrypto/bls0chain_test.go
cnlangzi: leave it first. we still need to migrate HermuiBLS to Miracl. Let's clean all of them once it is migrated.

Makefile
cnlangzi: gosdk-test and wasm-test can't run on same environment. so there is a new ci job named `wasm-test` on github action.

PR #295, opened 2021-11-11
fix(zcncrypto): fixed SingatureScheme interface

cnlangzi:
- refactored code to work with SignatureScheme instead of BLS0ChainScheme


rizary (approved):
lgtm

PR #303, opened 2021-11-24
add Available feature to the AuthTicket

rizary:
Adding feature to add `available-after` in sdk based on https://github.com/0chain/blobber/issues/439


moldis (approved):
Looks good

PR #309, opened 2021-12-01
feat: get wasm sdk worked with js-client-sdk and block-explorer

### LINKED ISSUE #294: migrate herumi bls for wasm
We need get `bls` worked with wasm sdk. there are 3 options
- migrate bls from herumi to miracl:
`aggregate signature and threshold signature are not built-in. we have to implement them`
- migrate bls from native herumi-bls to standalone  hermui-bls wasm
`standalone wasm is not provided by herumi, we need call emcc to build it by ourself `
- migrate bls from native herumi-bls to web hermui-bls wasm (bls_c.js + bls.js)
`aggregate and threshold are missing on official bls.js, we need expose them by ourself`

cnlangzi:


core/zcncrypto/helper.go
```diff
@@ -24,6 +25,6 @@ func MiraclToHerumiPK(pk string) string {
 	n3 := pk[(66 + 64):(66 + 64 + 64)]
 	n4 := pk[(66 + 64 + 64):(66 + 64 + 64 + 64)]
 	var p bls.PublicKey
-	p.SetHexString("1 " + n2 + " " + n1 + " " + n4 + " " + n3)
+	p.SetHexString("1 " + n2 + " " + n1 + " " + n4 + " " + n3) //nolint
```
rizary: I don't think we need the `MiraclToHerumiPK` function after this PR, no?

wasmsdk/jsbridge/input.go
```diff
@@ -158,3 +162,20 @@ func jsValueToFloat64(jv js.Value) reflect.Value {
 
 	return reflect.ValueOf(i)
 }
+
+func jsValueToStringSlice(jv js.Value) reflect.Value {
+	var list []string
+
+	if jv.Truthy() {
```
rizary: `Truthy` didn't cover if js.Value is type of Array specifically. Hence, we need to call `isArray` function to be precised.

wasmsdk/jsbridge/input_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build js && wasm
 // +build js,wasm
 
 package jsbridge
```
rizary, line 4: need to add more test in the TestBinder for javascript array

rizary (changes requested):
We still need to think custom type for the binder.

But if you want to do it in separate PR, that's fine for me.

cnlangzi:
> We still need to think custom type for the binder.
> 
> But if you want to do it in separate PR, that's fine for me.

Yes. new custom binder will be added if it necessary in the future. for now. it works enough for our wasm sdk

rizary (approved):
lgtm

PR #324, opened 2021-12-09
make default auth ticket  expiration to never expire

### LINKED ISSUE #455: For private share expiry_at should support 0 value
I case if we set expiry_at to 0 - file should be available infinitely

rizary:
fix https://github.com/0chain/blobber/issues/455


cnlangzi (approved):
LGTM

PR #328, opened 2021-12-14
Revert "add Available feature to the AuthTicket"

cnlangzi:
Reverts 0chain/gosdk#303


cnlangzi:
@Rizary  let's revert this feature first. It seems break system_test.

Issue #333, opened 2021-12-21
Get allocation object info


lpoli:
After gosdk is initialized, we can get `allocationObj` as `allocation, err := sdk.GetAllocation(allocationID)`. What this function does is, it gets allocation information from smart contract and create `sdk.Allocation` object and it will initialize allocation object as well which re-makes channels(upload, download, commit, etc.).

Above is alright if we use zboxcli type of application which always use fresh allocation object and operates on it. When other application like; 0fs, s3-migration, 0s3-server are to use gosdk then it is better to get updated allocation information. We cannot use `sdk.GetAllocation(allocationID)` as this would re-initialized above mentioned channels and other fields as well. For the running applications we need similar function as `sdk.GetAllocation` that doesn't initialize existing allocation fields.

PR #334, opened 2021-12-21
Feature/get allocation info

lpoli:
https://github.com/0chain/gosdk/issues/333

### DIFF
--- a/zboxcore/sdk/sdk.go
+++ b/zboxcore/sdk/sdk.go
@@ -781,6 +781,46 @@ func GetAllocation(allocationID string) (*Allocation, error) {
 	return allocationObj, nil
 }
 
+func GetAllocationUpdates(allocation *Allocation) error {
+	if allocation == nil {
+		return errors.New("allocation_not_initialized", "")
+	}
+
+	params := make(map[string]string)
+	params["allocation"] = allocation.ID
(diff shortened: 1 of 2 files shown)


zboxcore/sdk/sdk.go
```diff
@@ -781,6 +781,46 @@ func GetAllocation(allocationID string) (*Allocation, error) {
 	return allocationObj, nil
 }
 
+func GetAllocationUpdates(allocation *Allocation) error {
```
moldis, line 784: @lpoli May I know where is it used? And why it's returning an error instead of an Allocation object?

lpoli: It is actually needed when running tools like s3-migration, 0fs. These tools takes first instance of allocation object which initializes in itself upload, download, commit, etc channels.
Now while uploading files its better to know in advance for example; used size.

Allocation can also be updated by user with zbox cli(`zbox updateallocation`) and that information might be relevant for other applications as well.

If only single process would communicate with an allocation then it is not required but there can be multiple processes communicating thus changing allocation.

Returning allocation object is not necessary because this function receives pointer of original allocation instance. Error is to notify caller of the occurred error.

moldis (approved):

PR #339, opened 2021-12-26
Feature/add system tests

stewartie4:


.github/workflows/tests.yml
```diff
@@ -40,6 +45,93 @@ jobs:
...
+          zdns_image: staging
+          explorer_image: latest
+          zproxy_image: staging
+          zsearch_image: staging
+          blobber_stake_image: latest
+
+      - name: "Run System tests"
+        uses: 0chain/actions/run-system-tests@master
+        with:
+          custom_go_sdk_version: ${{ env.CURRENT_BRANCH_HEAD }}
```
stewartie4, line 97: This will 
1. checkout zwalletcli and update it to use the gosdk version from the current branch
2. checkout zboxcli and update it to use the gosdk version from the current branch
3. run system tests 

If tests fail, or if staging CLI is not compatible with current SDK then manual system tests can be ran as in other repos

PR #340, opened 2021-12-26
add system tests

stewartie4:


cnlangzi:
please add `system-tests.yml` for trigger manually. sometimes a break change is introduced, we need migrate zbox code too. It doesn't work by upgrading gosdk version in go.mod only. custom zboxcli/zwalletcli branch may be configured for any break changes on gosdk.

stewartie4:
this will pass once *zwallet is upgraded to latest gosdk

cnlangzi (approved):
LGTM

PR #341, opened 2021-12-27
Fix gosdk issues that are raised by different linters

lpoli:
There were some inconsistencies and still are.
I fixed some issues raised by `golangci linter`

I've reduced code redundancy as:
`for true{}` to `for{}`.

Variables that are unexported and unused are also removed.
Possible context leak issue is also fixed

Some possible blocking by channel issue is also fixed.
There are lots of other issues; Please check with `golangci-lint run`. I could not decide in some issues.

### DIFF
--- a/zboxcore/client/entity.go
+++ b/zboxcore/client/entity.go
@@ -10,7 +10,7 @@ type SignFunc func(hash string) (string, error)
 
 type Client struct {
 	*zcncrypto.Wallet
-	signatureSchemeString string `json:"signature_scheme"`
+	signatureSchemeString string
 }
 
 var (
--- a/zboxcore/fileref/fileref.go
+++ b/zboxcore/fileref/fileref.go
@@ -57,21 +57,21 @@ func (a *Attributes) Validate() (err error) {
 }
 
 type FileRef struct {
-	Ref
-	CustomMeta          string          `json:"custom_meta"`
-	ContentHash         string          `json:"content_hash"`
-	MerkleRoot          string          `json:"merkle_root"`
-	ThumbnailSize       int64           `json:"thumbnail_size"`
-	ThumbnailHash       string          `json:"thumbnail_hash"`
-	ActualFileSize      int64           `json:"actual_file_size"`
-	ActualFileHash      string          `json:"actual_file_hash"`
-	ActualThumbnailSize int64           `json:"actual_thumbnail_size"`
-	ActualThumbnailHash string          `json:"actual_thumbnail_hash"`
-	MimeType            string          `json:"mimetype"`
-	EncryptedKey        string          `json:"encrypted_key"`
-	CommitMetaTxns      []CommitMetaTxn `json:"commit_meta_txns"`
-	Collaborators       []Collaborator  `json:"collaborators"`
-	Attributes          Attributes      `json:"attributes"`
+	Ref                 `mapstructure:",squash"`
+	CustomMeta          string          `json:"custom_meta" mapstructure:"custom_meta"`
+	ContentHash         string          `json:"content_hash" mapstructure:"content_hash"`
+	MerkleRoot          string          `json:"merkle_root" mapstructure:"merkle_root"`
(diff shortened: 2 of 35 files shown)


zboxcore/client/entity.go
```diff
@@ -10,7 +10,7 @@ type SignFunc func(hash string) (string, error)
 
 type Client struct {
 	*zcncrypto.Wallet
-	signatureSchemeString string `json:"signature_scheme"`
```
cnlangzi, line 13: I would suggest to leave `json` there. We don't know if it is used by anyone.

zboxcore/fileref/fileref.go
```diff
@@ -57,7 +57,7 @@ func (a *Attributes) Validate() (err error) {
 }
 
 type FileRef struct {
-	Ref                 `json:",squash"`
+	Ref
```
cnlangzi: I would suggest to leave `squash` there. Because we don't know if it is default on embedded struct in all mapstructure versions.

zboxcore/sdk/allocation_test.go
```diff
@@ -344,8 +344,10 @@ func TestAllocation_dispatchWork(t *testing.T) {
 		a.uploadChan <- &UploadRequest{file: []*fileref.FileRef{}, filemeta: &UploadFileMeta{}}
 	})
 	t.Run("Test_Cover_Download_Request", func(t *testing.T) {
+		ctx := context.Background()
 		go a.dispatchWork(context.Background())
-		a.downloadChan <- &DownloadRequest{}
+		ctx, ctxCncl := context.WithCancel(ctx)
```
cnlangzi:
```suggestion
ctx, ctxCncl := context.WithCancel(context.Background())
```

zcnbridge/keystore.go
```diff
@@ -39,7 +39,7 @@ func AccountExists(address string) bool {
 		return false
 	}
 
-	status, err := wallet.Status()
+	status, _ := wallet.Status()
```
cnlangzi, line 42: do you think it should return false if err is not nil?

cnlangzi (changes requested):

zboxcore/client/entity.go
lpoli: No. The field is not exported so it won't be used by json as well

zboxcore/fileref/fileref.go
lpoli: I checked with json tags and there is no squash option.
For mapstructure, it should have mapstructure in place of json, right?

zcnbridge/keystore.go
lpoli: I am not sure how it is defined. Wallet can have any status and be thought as existed.
Even the call to `wallet.Status()` is done to just print the information.

zboxcore/sdk/allocation_test.go
lpoli: Yes sure

lpoli: Done

zboxcore/fileref/fileref.go
cnlangzi: I think it is designed for mapstructure, not for json. so please double check it on validator.

cnlangzi: I would suggest to leave it if we can't know how it works on old code. I don't know all old code as well .

cnlangzi:
@lpoli  you have to check system_tests. some tests seems are broken.

lpoli:
> 

Found the issue:
```
config := &mapstructure.DecoderConfig{
			Metadata: &md,
			Result:   rootRef,
			TagName:  "json",
		}
```
This is done in multiple places. I shall fix this with appropriate tag for mapstructure.

cnlangzi:
> mapstructure

it says mapstructure works with `json` tag. i think it is ok?

lpoli:
@cnlangzi The issue regarding mapstructure is completely fixed.
Can you approve the PR?

cnlangzi (approved):
LGTM

PR #347, opened 2021-12-28
send zcn with description as correctly serialised json object

### LINKED ISSUE #346: Wrap send ZCN description in a json object before setting transaction_data
When sending ZCN, the transaction_data field is used to provide the transaction note/description.
There is now a requirement that this transaction_data should contain a valid json object

The SDK continues to passthrough the description text directly with no json wrapping causing the transaction to be rejected with invalid_transaction_data error.

Fix:
The SDK should wrap the "description" user input in a json object, then set the transaction_data field with this object.
eg.
transaction_data = "description"
becomes
transaction_data = {"note" : "description" }

stewartie4:


shaktals:
Cool, so, on the frontend we are now stringifying the note prop like:
`JSON.stringify({ note: "random text" })`
Wich would result in a string like:
`"{\"note\":\"random text\" }"`

Now that you are _marshalling_ the description parameter, does it mean that won't need to stringify the `note` prop anymore?

stewartie4:
@shaktals yes, you can pass the description in as a plain string "random text" - you won't need the wrapper object with the note field anymore. 
I was thinking of your guys and your "hack" hopefully fixing this in a single place helps with that, if not I can serialise at the CLI level and leave the SDK as-is but I think this is cleaner as mobile is affected too
Let me know which you prefer

shaktals:
Solving it as sdk level definitely sounds better than adding another hack on the CLI, besides the current on explorer :grinning:

cnlangzi:
We need double check it on sharder/miner. Because transaction_data is consumed by shader/miner too.

stewartie4:
@cnlangzi I'm not sure wha you mean - this txn is sent to the miner/sharder for processing like every other txn 
this only fix here is to ensure we are wrapping the description in a json object correctly rather than relying on the user constructing it

cnlangzi (approved):
approve it to fix `zwallet send` bug first.

shaktals:
To be precise on what's being done, we're creating a json object with the description string and then _stringifying_ it :)

stewartie4:
yeah stringily would be the commonly used js term due to the method name
java uses the term serialize or marshal and i have a java bias 😆

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | feature/fix-send-zcn  |
| gosdk  | feature/send-zcn-with-description-fix  |
| 0boxcli  | staging  |
| 0walletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

shaktals:
> yeah stringily would be the commonly used js term due to the method name java uses the term serialize or marshal and i have a java bias laughing

yeah, then you are using a java language paradigm on something called 'JavaScript Object Notation' :grin:

laughs apart, as I understand, you can serialize/marshal something into `json`, but it doesn't mean the result will be a _stringified_ json, right?

and that is the distinction I find important to make:
- a json object
- a stringified json object

because obviously, those are different things

PR #354, opened 2022-01-10
Add fix draft for encryption issue

lpoli:
This is a draft PR to address encryption issue with chunked upload.

Encryption increases file size by `16 bytes`. Its checkSums(message checksum and overall checksum) are `128 bytes` each separated by `,`. So total increase in block size if encryption is applied will be 273 bytes.

For the given chunkSize we should update chunkSize by adding this `Encryption overhead`.
So for example; if I have `4 bytes ` of data and allocation with `4 data-shards` then if encryption is not applied each shards will contain 1 byte of data. It can have chunkSize of 1 byte. But if encryption is applied, chunkSize should be 274 bytes and each shards will contain 274 bytes of data.
If chunksize is not updated to `encryption overhead` then `newChunkSize` will become negative.

FixedMerkleTree has calculation that panics if chunkSize is lesser than 1024 bytes. It should be fixed as well


cnlangzi:
@lpoli @guruhubb do you think it is better to limit chunksize that must be greater than 1KB?

cnlangzi:
@lpoli @guruhubb  do you think it is better to limit chunksize that must be greater than 1KB?

cnlangzi:
btw, this is not a simple encryption issue. it is `share` issue. I remember a file will be re-encrypted again if it is shared by auth ticket.   
I think download encrypted file works for owner. so it should be not a solution to change code in `ChunkedUpload`. as you see, many system tests fail now.

lpoli:
@cnlangzi If we put chunkSize limit to >= 1KB then what happens is data won't be distributed between blobbers properly.
If I have file size that is lesser than 1KB and if I apply 1KB chunkSize then `n` (data-shards+parity-shards) will have to store (n-1)KB padded bytes which is not efficient.

Its not an `share` issue. Actually Encryption with auth-ticket in blobber just considers 273 bytes over-head. We need to have it in chunked upload as well.

@cnlangzi We need to avoid this overheads. This is the cost that client should bear. So we should allow client to have total customizable chunkSize.

cnlangzi:
> @cnlangzi If we put chunkSize limit to >= 1KB then what happens is data won't be distributed between blobbers properly. If I have file size that is lesser than 1KB and if I apply 1KB chunkSize then `n` (data-shards+parity-shards) will have to store (n-1)KB padded bytes which is not efficient.
> 
> Its not an `share` issue. Actually Encryption with auth-ticket in blobber just considers 273 bytes over-head. We need to have it in chunked upload as well.
> 
> @cnlangzi We need to avoid this overheads. This is the cost that client should bear. So we should allow client to have total customizable chunkSize.

hmm, it is for ChallengeHash only, not for storage. I will check it again.

cnlangzi:
```
FixedMerkleTree has calculation that panics if chunkSize is lesser than 1024 bytes. It should be fixed as well
```
it is fixed on https://github.com/0chain/gosdk/pull/370

cnlangzi:
@lpoli  encryption header issues are fixed on https://github.com/0chain/gosdk/pull/373

cnlangzi:
it think all issues mentioned in this PR are fixed. 
about change `2KB to 273 bytes`, please confirm it from @guruhubb . I will update it if we all agreed to change it with less bytes to save space in chunk.

cnlangzi:
EncryptionHeaderSize is reduced with 256 on https://github.com/0chain/gosdk/pull/374. I think this PR can be closed. @lpoli

PR #357, opened 2022-01-12
Feature/transaction nonce

### LINKED ISSUE #356: [FEATURE] Transaction nonce
To be compatible with [changes in 0chain](https://github.com/0chain/0chain/issues/765), following changes should be implemented in gosdk:
1. client of gosdk should be able to provide nonce for every transaction he creates
2. If nonce is not provided (provided any number less than 1) gosdk should query `balance` from sharders, increment obtained nonce and use it for current transaction
3. Add nonce to transaction hash calculation

dabasov:
Added minimal implementation to support transaction nonce logic. 

- [ ]  Ensure system tests are passing. If not [Run them manually](https://github.com/0chain/blobber/actions/workflows/system_tests.yml) to check for any regressions :clipboard:
- [ ]  Merge your system tests PR to master BEFORE merging this PR

Associated PRs (Link as appropriate):
...


zcncore/wallet.go
```diff
@@ -627,6 +632,24 @@ func GetBalance(cb GetBalanceCallback) error {
...
+func GetNonce(cb GetNonceCallback) error {
+	err := checkConfig()
+	if err != nil {
+		return err
+	}
+	go func() {
+		value, info, err := getNonceFromSharders(_config.wallet.ClientID)
+		if err != nil {
+			Logger.Error(err)
+			cb.OnNonceAvailable(StatusError, 0, info)
```
peterlimg, line 672: Check if `cb` is nil to avoid potential panic perhaps.

peterlimg (changes requested):

zcncore/wallet.go
dabasov: fixed

kushthedude (commented):
Looks good to me

cnlangzi (dismissed):
LGTM

kushthedude:
Tests would fail due to changes of this new field, can we fix the system_tests & then merge this PR?

core/transaction/entity.go
```diff
@@ -44,6 +44,7 @@ type Transaction struct {
 	TransactionType   int    `json:"transaction_type"`
 	TransactionOutput string `json:"transaction_output,omitempty"`
 	TransactionFee    int64  `json:"transaction_fee"`
+	TransactionNonce  int64  `json:"transaction_nonce"`
```
peterlimg, line 46: Actually don't know why we have these stuttering field names such as Transaction.TransactionOutput, Transaction.TransactionFee, so as the nonce, we may need to change these later to Output, Fee, and Nonce.

zcncore/transaction.go
```diff
@@ -226,22 +228,47 @@ func (t *Transaction) completeVerifyWithConStatus(status int, conStatus Confirma
...
 	t.txnError = nil
 
+	nonce := t.txn.TransactionNonce
+	if nonce < 1 {
+		nonce = transaction.Cache.GetNextNonce(t.txn.ClientID)
+	} else {
+		transaction.Cache.Set(t.txn.ClientID, nonce)
+	}
+	t.txn.TransactionNonce = nonce
+
```
peterlimg: If the nonce was < 1, and we get value from cache, and update the transaction, shouldn't we sign this transaction again? maybe we should return error here instead of update the nonce. The caller should set the nonce and signed it correctly before submit.

zcncore/transactionauth.go
```diff
@@ -108,6 +112,14 @@ func (ta *TransactionWithAuth) sign(otherSig string) error {
 }
 
 func (ta *TransactionWithAuth) submitTxn() {
+	nonce := ta.t.txn.TransactionNonce
+	if nonce < 1 {
+		nonce = transaction.Cache.GetNextNonce(ta.t.txn.ClientID)
+	} else {
+		transaction.Cache.Set(ta.t.txn.ClientID, nonce)
+	}
+	ta.t.txn.TransactionNonce = nonce
```
peterlimg, line 121: The same problem as Transaction, return error instead of changing the Nonce.

peterlimg (changes requested):

core/transaction/cache.go
```diff
@@ -0,0 +1,143 @@
...
+	"strconv"
+	"sync"
+)
+
+const GET_BALANCE = `/v1/client/get/balance?client_id=`
+const consensusThresh = float32(25.0)
+
+//var defaultLogLevel = logger.DEBUG
+//var Logger logger.Logger
+var Cache *NonceCache
```
peterlimg, line 19: The cache would not last after a client command is executed, I'm wondering if this cache is necessary as basically each time when we run a command, this cache will be recreated

zcncore/transaction.go
dabasov: @peterlimg sorry, just found your comment. Do you mean that in case of transaction reuse we can get problems since nonce will be updated and signature won't?

peterlimg: @dabasov I'm not sure if this is caused by reusing transaction, but the issue is we change the `transaction.Nonce` in `submitTxn()` method but the signature is not updated. We'd not changing any fields when submitting a transaction, it may cause unexpected result.

peterlimg (changes requested):

core/transaction/cache.go
dabasov: this cache will be recreated on every command, correct. in this case client has own mechanics to manage nonces. this cache is used for blobber and other clients, who initializes gosdk once and use it during it's lifetime

zcncore/transaction.go
dabasov: It will be signed if signature is not provided and it is not provided normally, I guess the only way transaction is signed that it is reused. I think we can force resign here

zcncore/transactionauth.go
peterlimg, line 121: The t.submitTxn() below will force sign the transaction again, which will reset the signature added in line 135 below.

peterlimg (changes requested):

peterlimg (approved):
lgtm

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | feature/transaction-nonce  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | skip-challenge  |
| gosdk  | feature/transaction-nonce  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #359, opened 2022-01-18
Refactor/allow 0 parity

lpoli:
This PR is about allowing zero parity-shards to be able to operate on files. Storage smart contract has no restriction on it as well.
It Contains the following changes:
1.  Pre-calculate consensuses and put it in allocation object. This reduces frequent calculation of consensuses.
2.  Return `noBlobber` error only if there are actually 0 blobbers
3.  Consensus threshold and consensus required for ok is 100% if there are 0 parity-shards
4. Remove `DeleteFromBlobber` function that was mistakenly added to the repo.


zboxcore/sdk/allocation.go
```diff
@@ -638,12 +652,6 @@ func (a *Allocation) ListDirFromAuthTicket(authTicket string, lookupHash string)
 }
 
 func (a *Allocation) ListDir(path string) (*ListResult, error) {
-	consensusThresh := (float32(a.DataShards) * 100) / float32(a.DataShards+a.ParityShards)
-	fullconsensus := float32(a.DataShards + a.ParityShards)
-	return a.listDir(path, consensusThresh, fullconsensus)
-}
-
-func (a *Allocation) listDir(path string, consensusThresh, fullconsensus float32) (*ListResult, error) {
```
Kishan-Dhakan, line 646: IMHO do not remove this unexported function completely, keep this as ListDir() calling listDir()

lpoli: If you see then it makes no difference. 
The unexported listDir function just required consensuses but now it doesn't require to be passed as function parameter.
It is passed from allocation object

Kishan-Dhakan: It's very common in Go to return a private implementation of a public interface, even if there are no changes in params. That pattern is used throughout the stdlib (or even throughout gosdk).

cnlangzi (approved):

PR #366, opened 2022-02-03
Fix json unmarshalling into interface

lpoli:
This PR contains fix for json unmarshaling issue. Package `json` doesn't know which Type to use for some interface so it returns error.
I've provided respective Type and assigned the value.


cnlangzi:
@lpoli  it is fixed on https://github.com/0chain/gosdk/pull/369.

PR #450, opened 2022-05-02
Payload to be passed as interface{} into transaction instead of string

Dmdv:
### Changes
-

### Fixes
-

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other:

https://github.com/0chain/gosdk/issues/453
https://github.com/0chain/0chain/issues/1238
https://github.com/0chain/token_bridge_authserver/pull/13/


zcnbridge/transaction/functions.go
```diff
@@ -1,36 +1,29 @@
...
 	}
 
-	_, err = t.ExecuteSmartContract(
-		ctx,
-		zcncore.ZCNSCSmartContractAddress,
-		transaction.ZCNSC_ADD_AUTHORIZER,
-		string(buffer),
-		0,
-	)
+	err = t.callBack.waitCompleteCall(context.Background())
```
dabasov: add timeout here

zcnbridge/transaction/functions.go
```diff
@@ -1,36 +1,29 @@
...
 import (
 	"context"
-	"encoding/json"
 
-	"github.com/0chain/gosdk/core/transaction"
 	"github.com/0chain/gosdk/zcncore"
 )
 
-func (t *Transaction) AddAuthorizer(ctx context.Context, input *zcncore.AddAuthorizerPayload) error {
+func AddAuthorizer(input *zcncore.AddAuthorizerPayload) (*Transaction, error) {
```
dabasov: add context

zcnbridge/transaction/txn.go
```diff
@@ -70,7 +70,7 @@ func NewTransactionEntity() (*Transaction, error) {
 // ExecuteSmartContract executes function of smart contract with provided address.
 //
 // Returns hash of executed transaction.
-func (t *Transaction) ExecuteSmartContract(ctx context.Context, address, funcName, input string, val int64) (string, error) {
+func (t *Transaction) ExecuteSmartContract(ctx context.Context, address, funcName string, input interface{}, val int64) (string, error) {
```
dabasov, line 73: why do we need it and does not use appropriate funcs from gosdk, as for ZCNSCAddAuthorizer?

zcnbridge/transaction/functions.go
Dmdv: Added the context but the timeout is on the side of a client

zcnbridge/transaction/functions.go
Dmdv: added context

zcnbridge/transaction/txn.go
Dmdv: It actually does

core/conf/vars.go
```diff
@@ -48,10 +49,22 @@ func InitChainNetwork(n *Network) {
 
 	if network == nil {
 		network = n
+		normalizeURLs(network)
```
dabasov: can remove it

zcnbridge/transaction/functions.go
```diff
@@ -1,36 +1,29 @@
...
 
-	_, err = t.ExecuteSmartContract(
-		ctx,
-		zcncore.ZCNSCSmartContractAddress,
-		transaction.ZCNSC_ADD_AUTHORIZER,
-		string(buffer),
-		0,
-	)
+	err = t.callBack.waitCompleteCall(ctx)
+	t.Hash = t.scheme.Hash()
```
dabasov, line 23: is it really needed?

core/conf/vars.go
Dmdv: This is actually a fix

zcnbridge/transaction/functions.go
Dmdv: Yes, required for further transaction verification

zcnbridge/transaction/txn.go
```diff
@@ -84,22 +86,36 @@ func (t *Transaction) ExecuteSmartContract(ctx context.Context, address, funcNam
...
 	}
 
 	return t.Hash, nil
 }
 
-// Verify checks including of transaction in the blockchain.
-func (t *Transaction) Verify(ctx context.Context) error {
-	const errCode = "transaction_verify"
+func (t *Transaction) GetTransactionHash() string {
+	var txnout map[string]json.RawMessage
```
peterlimg: The transaction hash should exist once the transaction is created and signed, we don't need to parse the transaction result to get it.

peterlimg (dismissed):

zcnbridge/transaction/txn.go
Dmdv: I've removed it

zcnbridge/transaction/functions.go
dabasov: we should avoid using background context, add default timeout context

zcnbridge/transaction/functions.go
dabasov: but Transaction has 2 different hash fields, one in t.scheme and another in t. 
Why we didn't embed t.scheme in transaction, so we could remove hash field duplication and use something like this?
```go
func (t *Transaction) Hash() string {
    return t.scheme.Hash()
 
}
```

peterlimg (approved):
lgtm

PR #457, opened 2022-05-09
Mobile build pipeline

moldis:


.github/workflows/build-mobile.yml
```diff
@@ -0,0 +1,151 @@
...
+        uses: actions/checkout@v2
+
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v3
+        with:
+          go-version: ^1.18
+
+      - name: Install Gomobile
+        run: |
+          go install golang.org/x/mobile/cmd/gomobile@latest
```
cnlangzi, line 36: `go get` is replaced with `go install`.  `go get` should be fine to remove.
```sugesstion
          go install golang.org/x/mobile/cmd/gomobile@latest
          go install golang.org/x/mobile/cmd/gobind@latest
          gomobile init
```

.github/workflows/build-mobile.yml
```diff
@@ -0,0 +1,151 @@
...
+          go install golang.org/x/mobile/cmd/gobind@latest
+          go get golang.org/x/mobile/cmd/gobind
+          go get golang.org/x/mobile/cmd/gomobile
+          gomobile init
+        env:
+          GOPROXY: https://proxy.golang.org,direct
```
cnlangzi: they are default. It is unnecessary to config them here.

.github/workflows/build-mobile.yml
```diff
@@ -0,0 +1,151 @@
...
+      - name: Upload Artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: zcncore-ios
+          path: zcncore-ios.tar.gz
+          retention-days: 5
+
+  build-android:
+    name: Build-android
+    runs-on: ubuntu-latest
```
cnlangzi: please use `[selft-hosted, docker-build]` as per devops team ask.

.github/workflows/build-mobile.yml
```diff
@@ -0,0 +1,151 @@
...
+          go install golang.org/x/mobile/cmd/gomobile@latest
+          go install golang.org/x/mobile/cmd/gobind@latest
+          go get golang.org/x/mobile/cmd/gobind
+          go get golang.org/x/mobile/cmd/gomobile
+          gomobile init
+        env:
+          GOPROXY: https://proxy.golang.org,direct
+          GO111MODULE: "on"
+
+      - name: Restore Go mod cache
```
cnlangzi: it slow build speed when job is running on `self-hosted`. it is better to remove it.

.github/workflows/build-mobile.yml
```diff
@@ -0,0 +1,151 @@
...
+          tag: ${{ steps.vars.outputs.tag }}
+          overwrite: true
+          file_glob: true
+
+      - name: Upload Artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: zcncore-android
+          path: zcncore-android.tar.gz
+          retention-days: 5
```
cnlangzi: it is better to remove ` retention-days: 5` to keep artifacts forever. because we don't know when the developer need to download them.

.github/workflows/build-mobile.yml
```diff
@@ -0,0 +1,151 @@
...
+        run: make build-ios
+
+      - name: Compress
+        run: tar -czvf zcncore-ios.tar.gz out/iossdk/zcncore.xcframework
+
+      - name: Set output
+        id: vars
+        run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
+
+      - name: Upload binaries to release
```
cnlangzi, line 52: it should be enabled for tag only. so please add `if` condition here.

cnlangzi (changes requested):
@moldis do you think it is better to add `build-mobile.yml` on zboxmobile repo instead of gosdk?

moldis:
> @moldis do you think it is better to add `build-mobile.yml` on zboxmobile repo instead of gosdk?

This CI builds zcncore library for 0Wallet, zboxmobile builds 0box library for 0box Apps

cnlangzi (changes requested):
@moldis  it is not recommended to commit 3rd source code on our repo.  please install them on build stage if they are necessary. check docker build script as an example.

moldis:
> @moldis it is not recommended to commit 3rd source code on our repo. please install them on build stage if they are necessary. check docker build script as an example.

Agree, removed

go.mod
```diff
@@ -117,3 +117,4 @@ require (
 )
 
 //replace github.com/ethereum/go-ethereum => github.com/certifaction/go-ethereum v1.10.3-wasm
+//replace github.com/herumi/bls-go-binary => github.com/0chain/bls-go-library-zboxmobile v1.26.1
```
cnlangzi, line 120: in gosdk/0chain, it uses `github.com/herumi/bls-go-binary`, not a fork. is it a problem if it is upgraded with latest upstream? I would suggest `upstream first` if we don't make any fixing on our fork.

cnlangzi (changes requested):
A small suggestion for bls-go-binary.  
BTW: `build-android` fails on github runner. please check it.

go.mod
moldis: @cnlangzi For mobile we have to keep it from our forked repo, as there some changes. Legacy bls-go-binary is wrongly building binary for IOS and Mac

cnlangzi (approved):
LGTM

PR #459, opened 2022-05-13
Perf/improve blockchain confirm

### LINKED ISSUE #214: improve transaction confirmation performance from sharders
https://github.com/0chain/gosdk/blob/520f39c0fe70bd33a21c7eff1d269386ac653f86/zcncore/transaction.go#L909
```
go func() {
		for {
			// Get transaction confirmation from random sharder
			confirmBlock, confirmation, lfb, err := getTransactionConfirmation(1, t.txnHash)
			if err != nil {
				tn := int64(common.Now())
				Logger.Info(err, " now: ", tn, ", LFB creation time:", lfb.CreationDate)
				if util.MaxInt64(lfb.CreationDate, tn) < (t.txn.CreationDate + int64(defaultTxnExpirationSeconds)) {
					Logger.Info("falling back to ", getMinShardersVerify(), " of ", len(_config.chain.Sharders), " Sharders")
					confirmBlock, confirmation, lfb, err = getTransactionConfirmation(getMinShardersVerify(), t.txnHash)
				
```

A transaction is verified twice. But 2nd is same as 1st. It should be removed

https://github.com/0chain/gosdk/blob/520f39c0fe70bd33a21c7eff1d269386ac653f86/zcncore/transaction.go#L503
```
func getTransactionConfirmation(numSharders int, txnHash string) (*blockHeader, map[string]json.RawMessage, *blockHeader, error) {
	result := make(chan *util.GetResponse)
	defer close(result)

	numSharders = len(_config.chain.Sharders) // overwrite, use all
	queryFromSharders(numSharders, fmt.Sprintf("%v%v&content=lfb", TXN_VERIFY_URL, txnHash), result)
	maxConfirmation := int(0)
	txnConfirmations := make(map[string]int)
```
numSharders is overwritten by  len(_config.chain.Sharders) . 

They should be improved with min_confirmation and parallel requests like https://github.com/0chain/gosdk/issues/195

cnlangzi:
### Changes
- added health check before sending query to a random sharder
- support to cancel other query requests if it reaches consensus.
- added consensus checking on `latest_finalized_block`
### Fixes
- fixed #214 

### Tests

- zboxcli: https://github.com/0chain/zboxcli/pull/232
- zwalletcli: https://github.com/0chain/zwalletcli/pull/130

### DIFF
--- a/zcncore/transaction_query.go
+++ b/zcncore/transaction_query.go
@@ -228,17 +228,15 @@ func (tq *TransactionQuery) FromAny(ctx context.Context, query string) (QueryRes
 }
 
 // FromConsensus query transaction from all sharders whatever it is selected or offline in previous queires, and return consensus result
-func (tq *TransactionQuery) FromConsensus(ctx context.Context, query string, handle QueryResultHandle) (*QueryResult, error) {
+func (tq *TransactionQuery) FromConsensus(ctx context.Context, query string, handle QueryResultHandle) error {
 	if tq == nil || tq.max == 0 {
-		return nil, ErrNoAvailableSharders
+		return ErrNoAvailableSharders
 	}
 
 	urls := make([]string, 0, tq.max)
 	for _, host := range tq.sharders {
 		urls = append(urls, tq.buildUrl(host, query))
 	}
-
-	var result *QueryResult
 
 	r := resty.New()
 	r.DoGet(ctx, urls...).
@@ -257,7 +255,7 @@ func (tq *TransactionQuery) FromConsensus(ctx context.Context, query string, han
 
 			if handle != nil {
 				if handle(res) {
-					result = &res
+
 					cf()
 				}
 			}
@@ -267,23 +265,19 @@ func (tq *TransactionQuery) FromConsensus(ctx context.Context, query string, han
 
 	r.Wait()
 
-	if result == nil {
-		return nil, ErrNoConsensus
-	}
-
-	return result, nil
+	return nil
 }
 
 // GetFastConfirmation get txn confirmation from a random online sharder
-func (tq *TransactionQuery) GetFastConfirmation(ctx context.Context, txnHash string) (*blockHeader, map[string]json.RawMessage, blockHeader, error) {
+func (tq *TransactionQuery) GetFastConfirmation(ctx context.Context, txnHash string) (*blockHeader, map[string]json.RawMessage, *blockHeader, error) {
 	var confirmationBlockHeader *blockHeader
 	var confirmationBlock map[string]json.RawMessage
-	var lfbBlockHeader blockHeader
+	var lfbBlockHeader *blockHeader
 
 	// {host}/v1/transaction/get/confirmation?hash={txnHash}&content=lfb
 	result, err := tq.FromAny(ctx, tq.buildUrl("", TXN_VERIFY_URL, txnHash, "&content=lfb"))
 	if err != nil {
-		return nil, nil, lfbBlockHeader, err
+		return nil, nil, nil, err
 	}
 
 	if result.StatusCode == http.StatusOK {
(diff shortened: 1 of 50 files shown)


service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | perf/improve_blockchain_confirm  |
| zboxcli  | perf/improve_blockchain_confirm  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

zcncore/transaction_query.go
```diff
@@ -0,0 +1,401 @@
...
+
+			// it is offline, try next one
+			continue
+		}
+
+		return host, nil
+	}
+}
+
+// FromAny query transaction form any sharder that is not selected in previous queires. use any used sharder if there is not any unused sharder
```
peterlimg: form -> from

zcncore/transaction_query.go
```diff
@@ -0,0 +1,401 @@
...
+
+			if resp != nil {
+				res.StatusCode = resp.StatusCode
+				Logger.Debug(string(respBody))
+			}
+
+			if handle != nil {
+				if handle(res) {
+
+					cf()
```
peterlimg, line 262: Seems the calling of cf() actually does not cancel anything. All requests will be sent out anyway. Please check the r.Wait() below. Otherwise the function name should be FromAll().

zcncore/transaction_query.go
```diff
@@ -0,0 +1,401 @@
...
+			if txnConfirmations[cfmBlockHeader.Hash] > maxConfirmation {
+				maxConfirmation = txnConfirmations[cfmBlockHeader.Hash]
+
+				if maxConfirmation >= numSharders {
+					confirmationBlockHeader = cfmBlockHeader
+					confirmationBlock = cfmBlock
+
+					// it is consensus by enough sharders, and latest_finalized_block is valid
+					// return true to cancel other requests
+					return true
```
peterlimg, line 380: Due to the FromConsensus() issue above, even we return true here, it won't stop sending requests

peterlimg (changes requested):

zcncore/transaction_query.go
cnlangzi: `ctx, cf := context.WithCancel`, so cf will not stop to send requests, but it will send cancel signal.

zcncore/transaction_query.go
cnlangzi: It doesn't stop sending requests. it sends cancel signal to requests. I will save some time to `Wait`. Because some shader might response slowly. It is a performance issue on `Wait`. cancel signal  is sent to fix it.

zcncore/transaction_query.go
cnlangzi: fixed

zcncore/transaction_query.go
```diff
@@ -0,0 +1,401 @@
...
+
+		Logger.Error("round info parse error.", err)
+		return nil, nil, nil, err
+
+	}
+
+	return nil, nil, nil, thrown.Throw(ErrTransactionNotFound, strconv.Itoa(result.StatusCode))
+}
+
+func (tq *TransactionQuery) GetConsensusConfirmation(ctx context.Context, numSharders int, txnHash string) (*blockHeader, map[string]json.RawMessage, *blockHeader, error) {
```
peterlimg, line 321: This method won't stop until all the tq.sharders have been checked, we will need to fix the cancel function in resty.DoGet() and Wait() methods. Otherwise the Consensus here would just not make sense.

peterlimg (changes requested):

zcncore/transaction_query.go
cnlangzi: fixed

peterlimg (approved):
lgtm

PR #466, opened 2022-05-26
Fix/consensus bug

cnlangzi:
### Changes
-

### Fixes
- fixed consensus bug: it fails to get info if 1/2 sharders are down , and 1st query result from a down sharder.

### Tests

- zwalletcli: https://github.com/0chain/zwalletcli/pull/132

### DIFF
--- a/zcncore/transaction_query.go
+++ b/zcncore/transaction_query.go
@@ -250,11 +250,13 @@ func (tq *TransactionQuery) FromAll(ctx context.Context, query string, handle Qu
 				Error:      err,
 				StatusCode: http.StatusBadRequest,
 			}
-			Logger.Debug(req.URL.String() + " " + resp.Status)
 
 			if resp != nil {
 				res.StatusCode = resp.StatusCode
+				Logger.Debug(resp.Status + ": " + req.URL.String())
 				Logger.Debug(string(respBody))
+			} else {
+				Logger.Debug(req.URL.String())
 			}
 
 			if handle != nil {
--- a/zcncore/transaction_query.go
+++ b/zcncore/transaction_query.go
@@ -441,7 +441,7 @@ func (tq *TransactionQuery) GetInfo(ctx context.Context, query string) (*QueryRe
 	}
 
 	if consensusesResp.StatusCode != http.StatusOK {
-		return nil, errors.New("zcn: fails to get info " + string(consensusesResp.Content))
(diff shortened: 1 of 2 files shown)


service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | fix/consensus_bug  |
| zboxcli  | staging  |
| zwalletcli  | fix/consensus_bug  |
| 0chain  | staging  |
| blobber  | fix/upgrade_deps_0517  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

zcncore/transaction_query.go
```diff
@@ -399,3 +402,51 @@ func (tq *TransactionQuery) GetConsensusConfirmation(ctx context.Context, numSha
...
+	err := tq.FromAll(ctx, query,
+		func(qr QueryResult) bool {
+			if qr.StatusCode != http.StatusOK {
+				return false
+			}
+
+			hash := fnv.New64a()
+			hash.Write(qr.Content)
+			h := hash.Sum64()
+			consensuses[h]++
```
peterlimg: Sharders could be on  different latest finalized block when they receive the request, hence will give different results even they are all working well and synced. We may need to return the latest finalized block info, such as round number and block hash in each response from sharders, so clients can determinte which one can be used. Before that, comparing the content hash may not work.

peterlimg (changes requested):

zcncore/transaction_query.go
cnlangzi: @peterlimg  it is fixing bug on https://github.com/0chain/gosdk/blob/a8ac8468109d46a561df4639cdf5bf20d42a3c87/zcncore/wallet.go#L820.  should `GetInfo` be executed with the same logic as `verify`?

cnlangzi: there are two sharders (01 and 02). in old `getInfoFromSharders`, it will always fails when sharder01 is down, and sharder02 works. Because sharder01 gives 502 first, and sharder02 gives 200. so resultMap[502] and resultMap[200] are both 1. but 502 comes first. so it is last result.

peterlimg (approved):
lgtm

PR #472, opened 2022-05-31
0NFT SDK

Dmdv:
### Changes
-

### Fixes
-

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


znft/keystore.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
+)
+
+// ListStorageAccounts List available accounts
+func ListStorageAccounts(homedir string) []common.Address {
+	keyDir := path.Join(homedir, WalletDir)
+	ks := keystore.NewKeyStore(keyDir, keystore.StandardScryptN, keystore.StandardScryptP)
+	config := &accounts.Config{InsecureUnlockAllowed: false}
+	am := accounts.NewManager(config, ks)
```
dabasov: next func ``DeleteAccount`` uses the same 4 lines of code, can move it to another function or pass reference to Manager, make it struct instance probably

dabasov (dismissed):

znft/keystore.go
Dmdv: Refactored

znft/example/main.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/0chain/gosdk/znft"
+)
+
+var mnemonic string
+
+func main() {
+	mnemonic = "float actor page tip vital solution knock agent cage grant size funny"
+
+	// Create this configuration
+	cfg := &znft.Configuration{
```
dabasov, line 19: is this configuration hardcoded?

dabasov (changes requested):

znft/example/main.go
Dmdv: Because this is an example.
Configuration is set up with cmd args @dabasov

dabasov (approved):
lgtm

znft/keystore.go
```diff
@@ -0,0 +1,147 @@
...
+	keyDir := path.Join(homedir, WalletDir)
+	ks := keystore.NewKeyStore(keyDir, keystore.StandardScryptN, keystore.StandardScryptP)
+
+	// 2. Init wallet
+	wallet, err := hdw.NewFromMnemonic(mnemonic)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to import from mnemonic")
+	}
+
+	pathD := hdw.MustParseDerivationPath("m/44'/60'/0'/0/0")
```
peterlimg, line 120: Comments the bip44 wallet path params or links to the bip44 wiki page so that caller understand what does that mean.

znft/ethereum_wallet.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/ethereum/go-ethereum/accounts"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/accounts/keystore"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/pkg/errors"
+)
+
+func (conf *Configuration) CreateEthClient() (*ethclient.Client, error) {
```
peterlimg, line 18: We'd better not creating methods on configuration, instead, we should pass the config as param, so as all the other following methods. Also, we can pass necessary params only, so for the `CreateEthClient()`, we will only need to pass the `EthereumNodeURL`. For the `createSignedTransactionFromKeyStore()` we will only need to pass the password and value, etc.

znft/ethereum_wallet.go
```diff
@@ -0,0 +1,93 @@
...
+func (conf *Configuration) createSignedTransactionFromKeyStore() *bind.TransactOpts {
+	var (
+		signerAddress = common.HexToAddress(conf.WalletAddress)
+		password      = conf.VaultPassword
+		value         = conf.Value
+	)
+
+	client, err := conf.CreateEthClient()
+	if err != nil {
+		Logger.Fatal(errors.Wrap(err, "failed to create ethereum client"))
```
peterlimg, line 35: Let's return or exit with error explicitly (all the places below as well) as if any chance the Logger level is < FATAL, the logger would do nothing, therefore the following code will be executed unexpectedly, and could be serious bug.

znft/ethereum_wallet.go
```diff
@@ -0,0 +1,93 @@
...
+	ks := keystore.NewKeyStore(keyDir, keystore.StandardScryptN, keystore.StandardScryptP)
+	signer := accounts.Account{
+		Address: signerAddress,
+	}
+	signerAcc, err := ks.Find(signer)
+	if err != nil {
+		Logger.Fatal(errors.Wrapf(err, "signer: %s", signerAddress.Hex()))
+	}
+
+	chainID, err := client.ChainID(context.Background())
```
peterlimg, line 48: We'd better not use context.Background(), instead, let's use context passed by the user, i.e add `ctx context.Context` param to the `createSignedTransactionFromKeyStore` function, so as other functions below require context.

znft/keystore.go
Dmdv: Fixed in 432ae2654218b5007d6fe8ac1aefdbc28bd67c1f

znft/ethereum_wallet.go
Dmdv: Updated in be39946fd2c31c03f4a392442e4151b601b11b16

znft/ethereum_wallet.go
Dmdv: Fixed in 6a4ac251daacdd8e99308f388575ea47c0b5bd5b

znft/ethereum_wallet.go
Dmdv: Fixed in 6530477dc45ee27386c2d579e3feca85c1084d4c

PR #484, opened 2022-06-09
Feat/build zcncore for mobile apps

cnlangzi:
### Changes
- added mobile support on zcncore/TransactionSchema

### Fixes
- 

### Tests

-


zcncore/smartcontract_executor_mobile.go
```diff
@@ -0,0 +1,31 @@
+//go:build mobile
+// +build mobile
+
+package zcncore
+
+type SmartContractExecutor interface {
+	// ExecuteSmartContract implements wrapper for smart contract function
+	ExecuteSmartContract(address, methodName string, input string, val uint64) error
```
peterlimg, line 10: Why do we differ the signature, if this breaks the mobile, then we should let the mobile code updated to adjust the changes.

peterlimg (changes requested):

zcncore/smartcontract_executor_mobile.go
cnlangzi: At present, only a subset of Go types are supported. https://pkg.go.dev/golang.org/x/mobile/cmd/gobind

cnlangzi: All exported symbols in the package must have types that are supported. Supported types include:

- Signed integer and floating point types.

- String and boolean types.

- Byte slice types. Note that byte slices are passed by reference,
  and support mutation.

- Any function type all of whose parameters and results have
  supported types. Functions must return either no results,
  one result, or two results where the type of the second is
  the built-in 'error' type.

- Any interface type, all of whose exported methods have
  supported function types.

- Any struct type, all of whose exported methods have
  supported function types and all of whose exported fields
  have supported types.

peterlimg: okay, got it.

zcncore/smartcontract_executor_mobile.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+func (t *Transaction) ExecuteSmartContract(address, methodName string, input string, val uint64) error {
+	err := t.createSmartContractTxn(address, methodName, input, val)
```
peterlimg: I'm afraid we can not use the string `input` directly and pass it to the `t.createSmartContractTxn()`, we will need to decode it, otherwise the transaction input data would not work in 0chain. So do 

```go
scData := make(map[string]interface{})
err := json.Unmarshal([]byte(input), &scData)
t.createSmartContractTxn(address, methodName, scData, val)

```

So as it should be done to `TransactionAuth`

peterlimg (changes requested):

zcncore/smartcontract_executor_mobile.go
cnlangzi: In mobile apps, they pass json string as input. eg `"{name:\"facuet\"....}"`. it is unnecessary to decode/encode it again.

.github/workflows/build-mobile.yml
```diff
@@ -1,4 +1,4 @@
-name: build-mobile
+name: MobileSDKs
```
moldis: name seems not following standards

zcncore/smartcontract_executor_mobile.go
peterlimg: It's necessary,  string value of input will not work if it's not decoded into concrete type or the general json map. For example, the below two json encode data are totally different. 0chain smart contract executor will not decode string input.

```go
type Foo struct {
	Name string `json:"name"`
}

sd := SmartContractTxnData{
	Name:      "sc_foo",
	InputArgs: &Foo{Name: "foo"}}

d, _ := json.Marshal(&sd)
fmt.Println(string(d))

f := Foo{Name: "foo"}
fd, _ := json.Marshal(f)

sd1 := SmartContractTxnData{
	Name:      "sc_foo",
	InputArgs: string(fd),
}

d2, _ := json.Marshal(&sd1)
fmt.Println(string(d2))
	
	
# output
{"name":"sc_foo","input":{"name":"foo"}}
{"name":"sc_foo","input":"{\"name\":\"foo\"}"}
```

.github/workflows/build-mobile.yml
cnlangzi: reverted

zcncore/smartcontract_executor_mobile.go
cnlangzi: fixed

zcncore/smartcontract_executor_mobile.go
```diff
@@ -0,0 +1,33 @@
...
+
+import "encoding/json"
+
+type SmartContractExecutor interface {
+	// ExecuteSmartContract implements wrapper for smart contract function
+	ExecuteSmartContract(address, methodName string, input string, val uint64) error
+}
+
+func (t *Transaction) ExecuteSmartContract(address, methodName string, input string, val uint64) error {
+	err := t.createSmartContractTxn(address, methodName, json.RawMessage(input), val)
```
peterlimg, line 14: Doing the json.RawMessage(input) won't fix this issue I said before, this is just dong a []byte() conversion, but the input data structure that 0chain expects is a json object, or a map[string]interface{} that could represents the json object. Please decode the input to a map[string]interface{}, then pass it to the `t.createSmartContractTxn`.

peterlimg (changes requested):

zcncore/smartcontract_executor_mobile.go
cnlangzi: It will work. see demo https://go.dev/play/p/G8w3cO1DaPF

peterlimg: yes, didn't know the json.RawMessage could be used for this. 👍

zcncore/smartcontract_executor_mobile.go
```diff
@@ -11,7 +11,7 @@ type SmartContractExecutor interface {
 }
 
 func (t *Transaction) ExecuteSmartContract(address, methodName string, input string, val uint64) error {
-	err := t.createSmartContractTxn(address, methodName, json.RawMessage(input), val)
+	err := t.createSmartContractTxn(address, methodName, json.RawMessage([]byte(input)), val)
```
peterlimg: Seems  the []byte convert is unnecessary, the result is the same

peterlimg (changes requested):

zcncore/smartcontract_executor_mobile.go
cnlangzi: reverted

peterlimg (approved):
lgtm

PR #501, opened 2022-06-25
SDK functions for maintaining starred files on allocation

### LINKED ISSUE #468: Need to star and unstar file as well as get list of starred only files/folders
Related to https://github.com/0chain/blobber/issues/700

### LINKED ISSUE #700: Need to add extra field to files table "starred" true/false
We need it in frontend to star/unstar file and display in favorites view. Apply on both files and folders

avanaur:
### Changes

Added SDK functions for maintaining starred files on allocation.

SDK introduces a `starredFilesRegistry` that stores the full list of starred remote filepaths on a `.starred` registry file.

The provided functions are as follows.
- `func (a *Allocation) UpdateStarredFiles(paths []string) error` - uploads or updates the `.starred` file to contain the new list of starred files.
- `func (a *Allocation) GetStarredFiles() ([]string, common.Timestamp, error)` - retrieves the list of starred files from `.starred`
- `func (a *Allocation) GetStarredFilesLastUpdateTimestamp() (common.Timestamp, error)` - peeks the current `updated_at` timestamp of `.starred` file which is useful for knowing if last pulled copy is stale or outdated already.

### Fixes

Resolves https://github.com/0chain/blobber/issues/700
Resolves https://github.com/0chain/gosdk/issues/468

### Tests
- [x]  Ensure system tests are passing. If not [Run them manually](https://github.com/0chain/gosdk/actions/workflows/system_tests.yml) to check for any regressions :clipboard:

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


sculptex:
Just an idea, rather than calling functions and structs 'Starred' specifically, perhaps Starred / . starred could be an instance of a more generic named function / struct for easy re-use.

E.g. UpdateRegFiles() would have an instance created for starred with .starred Ref file, but Favorites could have instance with .favorites, .collection1 for a collection etc.

zboxcore/sdk/registry_file_manager.go
```diff
@@ -0,0 +1,203 @@
...
+	"sync"
+	"time"
+)
+
+// allocationFileStorer defines functions expected from Allocation for storing file.
+// This interface enables easy mocking for UT purposes.
+type allocationFileStorer interface {
+	ListDir(path string) (*ListResult, error)
+	DownloadFile(localPath string, remotePath string, status StatusCallback) error
+	StartChunkedUpload(workdir, localPath string, remotePath string, status StatusCallback, isUpdate bool, isRepair bool, thumbnailPath string, encryption bool) error
```
peterlimg, line 20: I would suggest to define a struct for all of those params, they are too long for a method and hard to read.

zboxcore/sdk/registry_file_manager.go
```diff
@@ -0,0 +1,203 @@
...
+	fileStorer       allocationFileStorer
+}
+
+var _ registryFileManager = &registryFileStore{}
+
+func newRegistryFileManager(a *Allocation, registryFilePath string) *registryFileStore {
+	return &registryFileStore{
+		registryFilePath: registryFilePath,
+		allocation:       a,
+		fileStorer:       a,
```
peterlimg, line 42: If the allocation and filestorer are the same thing, perhaps we only need one of them

zboxcore/sdk/registry_file_manager.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+}
+
+// Update writes the registry file with data provided.
+// This overwrites the current copy stored.
+func (a *registryFileStore) Update(data []byte) error {
+	rf, err := a.getRegistryFile()
+	if err != nil {
+		return errors.New("update_registry_file_failed", "Failed to check existence of registry file: "+err.Error())
```
peterlimg: Do not capitalize error message, so please use "failed to check exitence of registry file..."

zboxcore/sdk/registry_file_manager.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// if no registry yet, upload new registry file.
+	// otherwise, update registry file.
+	isUpdate := rf != nil
+
+	tempLocal := filepath.Join(os.TempDir(), fmt.Sprintf("registerfile_ul_%d", time.Now().UnixNano()))
+
+	err = os.WriteFile(tempLocal, data, 0600)
+	if err != nil {
+		return errors.New("update_registry_file_failed", "Failed to create file locally for upload: "+err.Error())
```
peterlimg: As above

zboxcore/sdk/registry_file_manager.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	err = a.fileStorer.StartChunkedUpload(os.TempDir(), tempLocal, a.registryFilePath, nil, isUpdate, false, "", false)
+	if err != nil {
+		return errors.New("update_registry_file_failed", "Failed to upload registry file: "+err.Error())
```
peterlimg: As above

zboxcore/sdk/registry_file_manager.go
```diff
@@ -0,0 +1,203 @@
...
+	rf, err := a.getRegistryFile()
+	if err != nil {
+		return errors.New("update_registry_file_failed", "Failed to check existence of registry file: "+err.Error())
+	}
+
+	// if no registry yet, upload new registry file.
+	// otherwise, update registry file.
+	isUpdate := rf != nil
+
+	tempLocal := filepath.Join(os.TempDir(), fmt.Sprintf("registerfile_ul_%d", time.Now().UnixNano()))
```
peterlimg: Use time.Now().UTC()

zboxcore/sdk/registry_file_manager.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	return nil
+}
+
+// Get retrieves the registry file contents and its last updated timestamp.
+func (a *registryFileStore) Get() ([]byte, common.Timestamp, error) {
+	rfLastUpdateTime, err := a.GetLastUpdateTimestamp()
+	if err != nil {
+		return nil, common.Timestamp(0), errors.New("get_registry_file_failed", "Failed to check existence of registry file: "+err.Error())
```
peterlimg: As above

zboxcore/sdk/registry_file_manager.go
```diff
@@ -0,0 +1,203 @@
...
+
+	f, err := os.Open(tempLocal)
+	if err != nil {
+		return nil, common.Timestamp(0), errors.New("get_registry_file_failed", "Failed to open downloaded registry file: "+err.Error())
+	}
+
+	defer os.Remove(tempLocal)
+	defer f.Close()
+
+	bt, err := io.ReadAll(f)
```
peterlimg, line 108: ReadAll could cause OOM if the file is large, so perhaps we need to return an io.Reader to caller along with close function to close the reader.

zboxcore/sdk/registry_file_manager.go
```diff
@@ -0,0 +1,203 @@
...
+	lastUpdate, err := time.Parse(time.RFC3339, rf.UpdatedAt)
+	if err != nil {
+		return common.Timestamp(0), errors.New("get_last_update_timestamp_failed", "Failed to parse last updated timestamp of registry file: "+err.Error())
+	}
+
+	return common.Timestamp(lastUpdate.Unix()), nil
+}
+
+func (a *registryFileStore) getRegistryFile() (*ListResult, error) {
+	dir, err := a.fileStorer.ListDir(a.registryFilePath)
```
peterlimg, line 128: This ListDir will only return one file (the registry file) right? The method name is a little confusing, I would thought it would list all the files in a dir. Perhaps rename it to List and wraps it to call the ListDir to return the file.

zboxcore/sdk/registry_file_starred.go
```diff
@@ -0,0 +1,64 @@
...
+// It is up to the SDK clients to ensure that their copy of starred files is always up-to date (not stale) before
+// doing a full overwrite of the saved starred list through UpdateStarredFiles(). Peeking
+//
+// It is recommended to flush any update to the list of starred files as often as possible to avoid list being stale for too long.
+
+const StarredRegistryFilePath = `/.starred`
+
+// UpdateStarredFiles writes the provided full list of starred files through the registry.
+func (a *Allocation) UpdateStarredFiles(paths []string) error {
+	data := []byte(strings.Join(paths, "\n"))
```
peterlimg, line 23: We'd better organize the paths in the .starrted file in json or msgpack format. It is for protecting the .starred file from getting malformed by using file names that itself contains the "\n". The formant protocol will help us to protect the integrity of data.

peterlimg (changes requested):
Please update all the capatial error messages, thanks.

zboxcore/sdk/registry_file_manager.go
avanaur: The interface allocationFileStorer is just a fraction of what Allocation struct does. I've added this as an interface so I can create unit tests where those functions such as `ListDir()` is mocked.

There is no integration tests on gosdk, and this is the simplest way to test I found.

zboxcore/sdk/registry_file_manager.go
avanaur: Ah yeah, seems with the latest changes, I just need to retain the fileStorer.

zboxcore/sdk/registry_file_manager.go
avanaur: `fileStorer.ListDir()` here is referring to the `Allocation.ListDir()` function

https://github.com/0chain/gosdk/blob/d7645bf6f88b1e76ac06e59dccb37a81a1af3143/zboxcore/sdk/allocation.go#L654

I'll just rename `getRegistryFile() (*ListResult, error)` to be `getRegistryFileLastUpdate() (common.Timestamp, error)`. This way it does need to leak `*ListResult`. Only important for users of that method is to know when was the last file update.

zboxcore/sdk/registry_file_starred.go
avanaur: Originally the PR is saving the list of starred files in a json format. @lpoli commented to keep it paths instead line by line. 

I am okay to set it back to JSON. Unless @lpoli has concerns.

zboxcore/sdk/registry_file_manager.go
avanaur: Do you see this will be issue for registry file purposes? The idea with @sculptex (at least with starred registry) is the app would keep the list of starred files inmemory. As such, would not expect it to be a lot.

If however we still need to take that large file in consideration, I may need to change not just this download of registry file but also upload. Upload may need to accept an `io.Reader` which expects registry file in correct format already (JSON or lines of paths). Parsing of download into format would be responsibility of caller as well.

peterlimg: okay, if the registry file is not too big, we can ignore it atm.

zboxcore/sdk/registry_file_starred.go
peterlimg: @lpoli please comment

lpoli: @avanaur I could not find the thread we discussed about the format. 
I might have thought about making it simple.

Yes and .starred file being malformed is a valid thing.
Lets use json. And @avanaur sorry for the inconvenience

peterlimg (approved):
lgtm

zboxcore/sdk/registry_file_starred.go
peterlimg: @avanaur please update to use json

peterlimg (changes requested):
please fix lint errors and use json for .starred file

dabasov:
we are doing it on 0box now

PR #517, opened 2022-07-30
Use time.unix.nano as transaction CreationDate

dabasov:
Use time.unix.nano to differentiate transactions, now if we execute transactions in the same second they have the same created time


peterlimg:
0chain is using transaction.CreationDate in second, if we want to use nano seconds, perhaps we will need to update 0chain as well, otherwise the txn with nano seconds will always be rejected.

dabasov:
@peterlimg why do we have `CreationDate` in nano here? 
https://demo.0chain.net/sharder01/v1/screst/6dba10422e368813802877a85039d3985d96760ed844092319743fb3a76712d7/transactions
it is confusing...

the problem is simple, we have this validation
```
	if marker.Timestamp >= now {
		return fmt.Errorf("marker timestamped in the future: %v", marker.Timestamp)
	}
```
	
now is a `transaction.CreationDate`, marker was issued in the same second and this check fails. maybe it can be easier to simply accept equal Timestamps. I'll fix it in another PR and we can cancel this one

peterlimg:
fixed in https://github.com/0chain/0chain/pull/1588

PR #555, opened 2022-08-26
feat(playlist): integrated playlist api for wasm player

cnlangzi:
### Changes
- integrated new playlist api for wasm player to improve performance

### Fixes
-

### Tests

- blobber: https://github.com/0chain/blobber/pull/814


zboxcore/sdk/playlist.go
```diff
@@ -0,0 +1,190 @@
...
+	opts = append(opts, resty.WithRetry(resty.DefaultRetry))
+	opts = append(opts, resty.WithTimeout(resty.DefaultRequestTimeout))
+	opts = append(opts, resty.WithRequestInterceptor(func(req *http.Request) {
+		req.Header.Set("X-App-Client-ID", client.GetClientID())
+		req.Header.Set("X-App-Client-Key", client.GetClientPublicKey())
+
+		hash := encryption.Hash(alloc.ID)
+		sign, err := sys.Sign(hash, client.GetClient().SignatureScheme, client.GetClientSysKeys())
+		if err != nil {
+			logger.Logger.Error("playlist: ", err)
```
peterlimg: Should not we return error here?

peterlimg (changes requested):

zboxcore/sdk/playlist.go
cnlangzi: error is unsupported in current `WithRequestInterceptor`. do you think it should be changed?

```
// WithRequestInterceptor intercept request
func WithRequestInterceptor(interceptor func(req *http.Request)) Option {
	return func(r *Resty) {
		r.requestInterceptor = interceptor
	}
}
```

peterlimg: hmm, if the sign result is a must, then we should check and return if error happens

cnlangzi: @peterlimg  fixed

peterlimg: Seems the request will be sent out anyway, though the response will be ignored. So I would suggest to move the siging code out of the option function so that we can check and return early, we should only need to set the signature in the option function.

peterlimg (changes requested):

zboxcore/sdk/playlist.go
cnlangzi: fixed

peterlimg (approved):
lgtm

PR #565, opened 2022-09-05
Enhancement/verify download

lpoli:
# Note for reviewer
Directly looking in downloadworker.go file will be irritating. Here's how I pushed the changes:
1. First I created newdownloadworker.go file and used go-composition and added few fields and methods of its own
2. Tested if download works with newdownloadworker.go
3. Deleted `downloadworker.go` and renamed `newdownloadworker.go` to `downloadworker.go` so it is complete replacement rather than changes to file.
4. This branch is created from fix/upload-issue branch

### Changes
 This PR has optimized download and made download process way more readable. The following changes are made:

- Few fields are added to DownloadRequest structure
- Download process is divided into more readable functions
- Download is done with as much minimum blobbers as possible (always keep `data+1` downloads if parity is not 0.

### Fixes
-

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | fix/upload-issue  |
| gosdk  | enhancement/verify-download  |
| zboxcli  | enhancement/verify-download  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | enhancement/verify-download  |
| gosdk  | enhancement/verify-download  |
| zboxcli  | enhancement/verify-download  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | fix/upload-issue  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

sculptex:
Cool. Regarding data+1 blobbers, are they randomised in any way? And is there fallback included if any of those chosen blobbers fail to respond?

zboxcore/sdk/downloadworker.go
```diff
@@ -46,338 +45,494 @@ type DownloadRequest struct {
...
+			break
+		}
 
-	var encscheme encryption.EncryptionScheme
-	if len(req.encryptedKey) > 0 {
-		encscheme = encryption.NewEncryptionScheme()
-		encscheme.Initialize(client.GetClient().Mnemonic)
-		encscheme.InitForDecryption("filetype:audio", req.encryptedKey)
+		if failed > remainingMask.CountOnes() {
+			return nil, errors.New("download_failed", "")
```
peterlimg: Why we return error if failed > remainingMask. We could at least send request to those remainingMask blobbers.

BTW, we can give more meaningful error message instead of empty "".

zboxcore/sdk/downloadworker.go
```diff
@@ -46,338 +45,494 @@ type DownloadRequest struct {
...
+	shards [][][]byte) (zboxutil.Uint128, int, error) {
+
+	var remainingMask zboxutil.Uint128
+	activeBlobbers := mask.CountOnes()
+	if activeBlobbers < requiredDownloads {
+		return zboxutil.NewUint128(0), 0, errors.New("insufficient_blobbers",
+			fmt.Sprintf("Required downloads %d, remaining active blobber %d",
+				req.consensusThresh, activeBlobbers))
+	}
+	rspCh := make(chan *downloadBlock, req.consensusThresh)
```
peterlimg: The channel size should be the requiredDownloads or activeBlobbers. Suppose if the consensusThresh < activeBlobbers, then more number of requests would be sent out than the respCh size, therefore goroutine leak could happen.

zboxcore/sdk/downloadworker.go
```diff
@@ -46,338 +45,494 @@ type DownloadRequest struct {
...
-	if err != nil {
-		return nil, errors.Wrap(err, "encoder init error")
-	}
-	for blockNum := 0; blockNum < decodeNumBlocks; blockNum++ {
-		data, err := erasureencoder.Decode(shards[blockNum], decodeLen[blockNum])
+
+	var failed int
+	for i := 0; i < requiredDownloads; i++ {
+		result := <-rspCh
+		err := req.fillShards(shards, result)
```
peterlimg: check the error before passing the result to fillShards, it should not be the fillShards()'s responsibility to check the result error.

zboxcore/sdk/downloadworker.go
```diff
@@ -46,338 +45,494 @@ type DownloadRequest struct {
...
-			sys.Files.Remove(req.localpath)
-			req.statusCallback.Error(req.allocationID, remotePathCallback, OpDownload, errors.New("invalid_block_num", "start block should be less than end block"))
-		}
+	remotePathCB := req.remotefilepath
+	if remotePathCB == "" {
+		remotePathCB = req.remotefilepathhash
+	}
+	fRef, err := req.getFileRef(remotePathCB)
+	if err != nil {
+		logger.Logger.Error(err.Error())
```
peterlimg, line 322: We should return error to caller instead of just logging. The caller may rely on the result to decide whether it should continue or not.

zboxcore/sdk/downloadworker.go
```diff
@@ -46,338 +45,494 @@ type DownloadRequest struct {
...
-		cnt := startBlock
-		logger.Logger.Info("Downloading block ", cnt+1)
-		if (startBlock + numBlocks) > endBlock {
+		if startBlock+numBlocks > endBlock {
 			numBlocks = endBlock - startBlock
 		}
+		logger.Logger.Info("Downloading block ", startBlock+1, " - ", startBlock+numBlocks)
 
-		data, err := req.downloadBlock(cnt+1, int(numBlocks))
+		data, err := req.getBlocksData(startBlock+1, numBlocks, req.downloadMask, req.consensusThresh)
```
peterlimg: The downloadMask and threshold are all member of `req`, so we don't need to pass them as param to the method

zboxcore/sdk/downloadworker.go
```diff
@@ -46,338 +45,494 @@ type DownloadRequest struct {
...
+	}
+	if req.encryptedKey != "" {
+		req.initEncryption()
+	}
+
+	var downloaded int
+	startBlock, endBlock, numBlocks := req.startBlock, req.endBlock, req.numBlocks
+	// remainingSize should be calculated based on startBlock number
+	// otherwise end data will have null bytes.
+	remainingSize := size - startBlock*int64(req.effectiveChunkSize)
```
peterlimg, line 380: explicitly check if remainingSize is < 0 to avoid unexpected bug.

zboxcore/sdk/downloadworker.go
```diff
@@ -46,338 +45,494 @@ type DownloadRequest struct {
...
-			sys.Files.Remove(req.localpath) //nolint: errcheck
-			if req.statusCallback != nil {
-				req.statusCallback.Error(req.allocationID, remotePathCallback, OpDownload, errors.Wrap(err, "Write file failed"))
-			}
+			req.errorCB(errors.Wrap(err, "Write file failed"), remotePathCB)
 			return
 		}
 		downloaded = downloaded + int(n)
-		size = size - n
+		remainingSize -= n
```
peterlimg, line 419: Check if remainingSize < 0

zboxcore/sdk/downloadworker.go
```diff
@@ -46,338 +45,494 @@ type DownloadRequest struct {
...
+
+	if expectedHash != hash {
+		err = errors.New("merkle_root_mismatch", "File content didn't match with uploaded file")
+		req.errorCB(err, remotepathCB)
+		return
+	}
+	return nil
+}
+
+func (req *DownloadRequest) getFileHandler(remotePathCB string) (sys.File, error) {
```
peterlimg: Rename to openFile, the Handler may confuse users and would be considered as a http handler

peterlimg (changes requested):

zboxcore/sdk/downloadworker.go
lpoli: mask will have only the active set of blobbers. Any failing blobber in between process is removed as you can see with the defer function.

remainingMask is the mask of blobbers that are not yet requested. Now if one blobber fails we request one blobber from remainingMask. If `failed` number of blobbers are greater than blobbers in remainingMask then download will fail regardless.

Empty error: I will put a meaningful error message.

zboxcore/sdk/downloadworker.go
lpoli: Actually, we don't request more than consensusThresh. It will suffice. 
`requiredDownload` is always lesser than consensusThresh. So its enough to use `requiredDownload` instead of consensusThresh

zboxcore/sdk/downloadworker.go
lpoli: This check is unnecessary because `startBlock` and `endBlock` are calculated based on size to download.
It will only be used in:
```
n := int64(math.Min(float64(remainingSize), float64(len(
_, err = mW.Write(data[:n])
```

`remainingSize` won't get chance to be negative.

zboxcore/sdk/downloadworker.go
lpoli: `getFileHandler` should be clear enough I think.
If you insist, I will change it.

lpoli:
> Cool. Regarding data+1 blobbers, are they randomised in any way? And is there fallback included if any of those chosen blobbers fail to respond?

@sculptex It is not randomised. Also it is good to get from `data` shards as much as possible because data-reconstruction can then be avoided.
And yes there is fallback and the request is made to the remaining unused blobber.

zboxcore/sdk/downloadworker.go
peterlimg: Yes, please change it. We don't use `handler` in go often to represent a file struct. It might be common in Java, C/C++, but let's keep it as file as the go standard code does.

sculptex:
> > Cool. Regarding data+1 blobbers, are they randomised in any way? And is there fallback included if any of those chosen blobbers fail to respond?
> 
> @sculptex It is not randomised. Also it is good to get from `data` shards as much as possible because data-reconstruction can then be avoided. And yes there is fallback and the request is made to the remaining unused blobber.

So they will be sequenced in order, e.g. from data[1]-data[n]+parity[0] ?

This drains read pool for those blobbers and neglects remaining parity blobbers. The fair way would be random including all parity although, as I have expressed previously, user should have option to favour blobbers, like preferred blobbers model which should put those to front of random list (choose first n+1 from randomised preferred + randomized remainder).

This also overlaps and possibly interferes with tokenomics which consider blobber choice to be more evenly distributed.

illuyankas:
@lpoli 

Reconstruction time savings depends on the ecScheme

1- a 10+2 ecScheme takes longer to reconstruct than a 3+2 ecScheme. So as the number of data fragments increases chances are you will be falling back anyways to reconstruction and thus you'd actually be slower not faster overall as you now have a two step process.

2- Reconstruction speed depends on whether you are utilizing local parity or not. In a local parity scheme, the data fragments are logically divided into groups and there is one local parity per group. Rebuilding data for a single failure is much faster with a local parity as you need to read less fragments.

> > Cool. Regarding data+1 blobbers, are they randomised in any way? And is there fallback included if any of those chosen blobbers fail to respond?
> 
> 
> 
> @sculptex It is not randomised. Also it is good to get from `data` shards as much as possible because data-reconstruction can then be avoided.
> 
> And yes there is fallback and the request is made to the remaining unused blobber.

illuyankas:
@sculptex 

The effect on tokenomics will be minimal for large data fragment allocations as chances are there will be lots of fallbacks on parity blobbers anyways - BUT that's due an undesired causation, high fallbacks.

For small data fragment allocations - you the parity blobbers will be at a disadvantage. 

Random + preferred is better. There should always be a random portion imo even if there is preferred. 

> > > Cool. Regarding data+1 blobbers, are they randomised in any way? And is there fallback included if any of those chosen blobbers fail to respond?
> 
> > 
> 
> > @sculptex It is not randomised. Also it is good to get from `data` shards as much as possible because data-reconstruction can then be avoided. And yes there is fallback and the request is made to the remaining unused blobber.
> 
> 
> 
> So they will be sequenced in order, e.g. from data[1]-data[n]+parity[0] ?
> 
> 
> 
> This drains read pool for those blobbers and neglects remaining parity blobbers. The fair way would be random including all parity although, as I have expressed previously, user should have option to favour blobbers, like preferred blobbers model which should put those to front of random list (choose first n+1 from randomised preferred + randomized remainder).
> 
> 
> 
> This also overlaps and possibly interferes with tokenomics which consider blobber choice to be more evenly distributed.
> 
>

zboxcore/sdk/downloadworker.go
```diff
@@ -46,338 +45,526 @@ type DownloadRequest struct {
...
+					failedMu.Unlock()
+					req.removeFromMask(uint64(result.idx))
+					logger.Logger.Error(err)
+				}
+			}()
+			if !result.Success {
+				err = fmt.Errorf("Unsuccessful download. Error: %v", result.err)
+				return
 			}
+			err = req.fillShards(shards, result)
```
peterlimg, line 205: Seems the `req.fillShards()` is not thread safe. See the comment below.

zboxcore/sdk/downloadworker.go
```diff
@@ -46,338 +45,526 @@ type DownloadRequest struct {
...
-			if success >= req.datashards {
-				decodeNumBlocks = downloadChunks
-				break
+// fillShards will fill `shards` with data from blobbers that belongs to specific
+// blockNumber and blobber's position index in an allocation
+func (req *DownloadRequest) fillShards(shards [][][]byte, result *downloadBlock) (err error) {
+	for i := 0; i < len(result.BlockChunks); i++ {
+		var data []byte
+		if req.encryptedKey != "" {
+			data, err = req.getDecryptedData(result, i)
```
peterlimg, line 241: `req.getDecryptedData` is not thread safe, the `req.authTicket` is read without lock protection. Also, the `req.encScheme` seems not thread safe as well.

peterlimg (changes requested):

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | fix/upload-issue  |
| gosdk  | fix/upload-issue  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

zboxcore/sdk/downloadworker.go
lpoli: req.fillShards uses different slice index. So it is safe.

zboxcore/sdk/downloadworker.go
lpoli: @peterlimg I checked `zboxcore/encryption/pre.go file`.
None of the method except for initialization has assignment operation. I think we can have concurrent operation and still be safe.

I understand that if someone later on make changes in `pre.go` file then it has potential of creating issue.

What would you suggest?
Lock protection?

peterlimg (approved):
LGTM

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | enhancement/verify-download  |
| gosdk  | enhancement/verify-download  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

peterlimg:
@lpoli please fix the system test errors

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | enhancement/verify-download  |
| gosdk  | enhancement/verify-download  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | enhancement/verify-download  |
| gosdk  | enhancement/verify-download  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #579, opened 2022-09-19
fix(upload): fixed consensus issue on uploading

### LINKED ISSUE #545: Upload should be successful it data could be uploaded to `data+1` blobbers
Consider an allocation with `--data 2` and `--parity 2`. In this case if we could upload some data block to 3 i.e. `data-shards + 1` blobbers then it should be consensus OK and upload should be successful.

However in chunked upload, Any single failure will result in upload failure.

### LINKED ISSUE #578: Consensus on writemarker lock
We ask writemarker lock with the blobber. All the failing blobbers should be removed from the upload mask, and upload should be made only to the succeeding blobbers. Consensus should be calculated on the number of locks acquired.

**_Note: All the blobbers should be requested for the lock_**

We don't have any reason to use `resty` client. So modify it to commonly used http client from zboxcli utils.

cnlangzi:
### Changes
-

### Fixes
- fixed https://github.com/0chain/gosdk/issues/545
- fixed https://github.com/0chain/gosdk/issues/578

### Tests

- zboxcli: https://github.com/0chain/zboxcli/pull/298


zboxcore/sdk/chunked_upload.go
```diff
@@ -527,38 +509,69 @@ func (su *ChunkedUpload) processUpload(chunkStartIndex, chunkEndIndex int, fileS
...
 		}
 
-		go func(b *ChunkedUploadBlobber, buf *bytes.Buffer, form ChunkedUploadFormMetadata) {
+		go func(idx uint64, b *ChunkedUploadBlobber, buf *bytes.Buffer, form ChunkedUploadFormMetadata) {
 			err := b.sendUploadRequest(ctx, su, chunkEndIndex, isFinal, encryptedKey, buf, form)
 
 			// channel is not closed
 			if ctx.Err() == nil {
-				wait <- err
+				wait <- UploadError{
```
peterlimg, line 517: We should push to wait anyway. Imagine if the ctx is canceled early and we only pushed half  of the num to wait, then the for loop below that reading err from wait channel will get stuck.

zboxcore/sdk/chunked_upload.go
```diff
@@ -527,38 +509,69 @@ func (su *ChunkedUpload) processUpload(chunkStartIndex, chunkEndIndex int, fileS
...
+
+	return errConsensus
 
 }
 
 // processCommit commit shard upload on its blobber
 func (su *ChunkedUpload) processCommit() error {
+	shouldUnlock := true
+	err := su.writeMarkerMutex.Lock(context.TODO(), su.progress.ConnectionID)
+	defer func() {
```
peterlimg, line 563: Check err before the defer as if we lock failed, it is unnecessary to call unlock

zboxcore/sdk/writemarker_mutex.go
```diff
@@ -64,6 +64,28 @@ func CreateWriteMarkerMutex(client *client.Client, allocationObj *Allocation) (*
 	}, nil
 }
 
+//the minimum of M blobbers must accept the marker
+func (m *WriteMarkerMutex) getMinimumAccept(T int) int {
```
peterlimg: Let's not use the capital T as param variable, leave it for go generic using.

peterlimg (changes requested):

zboxcore/sdk/chunked_upload.go
cnlangzi: we still need try to unlock some of them. because some of blobbers might be locked. eg 1-3 are locked, but fails to lock 4,5,6 blobber.

zboxcore/sdk/writemarker_mutex.go
cnlangzi: fixed

zboxcore/sdk/chunked_upload.go
cnlangzi: fixed

zboxcore/sdk/chunked_upload.go
cnlangzi: and in WriteMarkerMutex, unlock will be skipped if a blobber is not locked.
```
		blobberUrl := builder.String()
		// only release lock on locked blobbers
		if m.lockedBlobbers != nil {
			if m.lockedBlobbers[b.ID] {
				urls = append(urls, blobberUrl)
			}
		} else { // Lock is not called here, try to release all blobbers
			urls = append(urls, blobberUrl)
		}
```

PR #580, opened 2022-09-20
added optimistic verification

dabasov:
### Changes
Implemented optimistic transaction verification

### Tests


core/transaction/utils.go
```diff
@@ -0,0 +1,400 @@
...
+			//need single valid confirmation
+			errs := r.First()
+			if len(errs) == 0 {
+				break L
+			}
+		}
+	}
+
+	if len(chain) == 0 {
+		return nil, errors.New("verify", "can't get confirmation after 10 retries")
```
peterlimg: It is 30 retries as show above, perhaps declare a variable for the retry times.

core/transaction/utils.go
```diff
@@ -0,0 +1,400 @@
...
+	toDelete := mappedSharders[url]
+	for i, s := range sharders {
+		if s == toDelete {
+			sharders = append(sharders[:i], sharders[i+1:]...)
+			break
+		}
+	}
+
+	//build blockchain starting from confirmation block
+	curRound := chain[0].Round
```
peterlimg, line 173: Let's move the confirmation code to a separate one to avoid making this function too long to read

core/transaction/utils.go
```diff
@@ -0,0 +1,400 @@
...
+		//iterate through all sharders sequentially to get next block
+		for {
+			next, err := rand.Next()
+			if err != nil {
+				return nil, errors.New("get_round_block", "can't get round block, blockchain might be stuck")
+			}
+
+			cur := sharders[next]
+			burl := fmt.Sprintf("%v/%v%v", cur, BLOCK_BY_ROUND_URL, curRound+1)
+			rb.DoGet(context.TODO(), burl)
```
peterlimg, line 230: Not sure if the resty instance could be reused without concurrent issue. We'd better check

core/transaction/utils.go
```diff
@@ -0,0 +1,400 @@
...
+			break
+		}
+
+		urls = []string{fmt.Sprintf("%v/%v%v", sharders[n], TXN_VERIFY_URL, txnHash)}
+
+	}
+
+	if numSuccess > 0 && numSuccess >= minNumConfirmation {
+		if retTxn == nil {
+			return nil, errors.Throw(ErrNoTxnDetail, strings.Join(msgList, "\r\n"))
```
peterlimg, line 426: msgList[0] is not set in this case, so perhaps `msgList[1:]`. Or just define the msgList normally as `msgList := make([]string, 0, numSharders)` and append it to TooLessConfirmation below as `append(TooLessConfirmErr, msgList...)`.

peterlimg (changes requested):

core/transaction/utils.go
dabasov: this is done also in previous implementation ``func VerifyTransaction(txnHash string, sharders []string) (*Transaction, error) {
``, so if we need to change it we need to do it in both places, can we do it in different issue?

core/transaction/utils.go
dabasov: this code is moved from and is not changed, so this might be not related to this issue, let's do it in different issue
```
func VerifyTransaction(txnHash string, sharders []string) (*Transaction, error) {
```

core/transaction/utils.go
peterlimg: Sure, let's do in another issue if needed.

peterlimg (approved):
lgtm

service-0chain:
| config  | value |
| tests skipped  | TRUE  |
| system-tests  | master  |
| gosdk  | feat/optimistic-verification  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #584, opened 2022-09-27
Refactor/improve delete perf

cnlangzi:
### Changes
-

### Fixes
- improved delete performance

### Tests

- zboxcli: https://github.com/0chain/zboxcli/pull/301


core/util/recover.go
```diff
@@ -0,0 +1,14 @@
+package util
+
+func WithRecover(fn func()) (success bool) {
+	defer func() {
+		if recover() != nil {
+			//recover panic from `send on closed channel`
```
peterlimg: It's kind of hacky to use recover() for this, instead, we should organize code better to avoid sending on closed channel. The principle is never closed the channel from the reader end, or close only when all writers have finish their work, which can easily achived by using the sync.WaitGroup.

zboxcore/sdk/deleteworker.go
```diff
@@ -26,109 +26,145 @@ type DeleteRequest struct {
...
-			refEntity, err := req.getObjectTreeFromBlobber(req.blobbers[blobberIdx])
+
+			fr, err := req.deleteFileFromBlobber(req.blobbers[blobberIdx])
 			if err == nil {
-				req.consensus.Done()
-				deleteMutex.Lock()
-				req.listMask |= (1 << uint32(blobberIdx))
-				objectTreeRefs[blobberIdx] = refEntity
-				deleteMutex.Unlock()
+				util.WithRecover(func() {
```
peterlimg: The util.WithRecover is not needed. We can define a `wg := sync.WaitGroup{}` and call `defer wg.Done()` in each goroutine function, then below the for loop, we call the `wg.Wait()` to wait for all the goroutines finish, till then should we start to read the `wait` channel, and we would not have panic issue.

peterlimg (changes requested):

core/util/recover.go
cnlangzi: removed

zboxcore/sdk/deleteworker.go
cnlangzi: fixed

peterlimg (approved):
lgtm

PR #615, opened 2022-10-20
wait some time before retrying downloading a file

fabioDMFerreira:
Repairing fails because there are multiple quick concurrent requests while the resource is locked.

This fix waits 1 second before retrying again.

### DIFF
--- a/zboxcore/sdk/blockdownloadworker.go
+++ b/zboxcore/sdk/blockdownloadworker.go
@@ -204,6 +204,7 @@ func (req *BlockDownloadRequest) downloadBlobberBlock() {
 				if bytes.Contains(respBody, []byte(LockExists)) {
 					zlogger.Logger.Debug("Lock exists error.")
 					shouldRetry = true
+					time.Sleep(time.Second * 1)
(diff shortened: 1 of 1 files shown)


lpoli (approved):

zboxcore/sdk/blockdownloadworker.go
```diff
@@ -204,6 +204,7 @@ func (req *BlockDownloadRequest) downloadBlobberBlock() {
 				if bytes.Contains(respBody, []byte(LockExists)) {
 					zlogger.Logger.Debug("Lock exists error.")
 					shouldRetry = true
+					time.Sleep(time.Second * 1)
```
cnlangzi, line 207: pleas use sys.Sleep, if not, main thread will be blocked in wasm sdk.

peterlimg: @cnlangzi if the time.Sleep will cause problem, please fix this in a new PR, thanks.

cnlangzi: @peterlimg it is fixed

PR #654, opened 2022-11-19
Feature: allow zcnbridge client to use custom wallet file name

### LINKED ISSUE #651: Feature: allow zcnbridge client to use custom wallet file name

YarikRevich:
Refers to issue #651


zcnbridge/config.go
```diff
@@ -221,22 +221,31 @@ func (b *BridgeOwner) ClientID() string {
...
+func loadWallet(homedir, fileName string) (*zcncrypto.Wallet, error) {
+	var walletPath string
+
+	if fileName != "" {
+		walletPath = path.Join(homedir, fileName)
+	} else {
+		walletPath = path.Join(homedir, ZChainWalletConfigName)
+	}
+
+	fmt.Println(walletPath)
```
peterlimg: Avoid print logs in sdk APIs or print with the `log` initialized so that the logs could be turned off when needed.

zcnbridge/config.go
```diff
@@ -221,22 +221,31 @@ func (b *BridgeOwner) ClientID() string {
...
+func SetupBridgeOwnerSDK(cfg *BridgeSDKConfig, walletFile string) *BridgeOwner {
 	log.InitLogging(*cfg.Development, *cfg.LogPath, *cfg.LogLevel)
-	bridgeOwner := CreateBridgeOwner(initBridgeConfig(cfg))
+	bridgeOwner := CreateBridgeOwner(initBridgeConfig(cfg), walletFile)
 	return bridgeOwner
 }
 
-func loadWallet(homedir string) (*zcncrypto.Wallet, error) {
-	walletPath := path.Join(homedir, ZChainWalletConfigName)
+func loadWallet(homedir, fileName string) (*zcncrypto.Wallet, error) {
```
peterlimg: Use optional params for the fileName so that the API changes would not affect exist code, i.e `loadWallet(homeDir)` can still work by default, and use `loadWallet(homeDir, walletFile)` if need to load specific wallet.

peterlimg (dismissed):

zcnbridge/config.go
YarikRevich: A good idea, thanks

zcnbridge/config.go
YarikRevich: Forgot to remove

dabasov (approved):
lgtm

PR #697, opened 2022-12-22
Feature: add a method to check if ethereum transaction is verified

### LINKED ISSUE #691: add method to verify eth transactions
we need to verify eth transaction from transaction hash to send receipts to bolt users
for that 0box needs to verify the txhash

YarikRevich:


dabasov (changes requested):
we better have simple rest call, since EthClient requires some additional resources

dabasov (approved):
lgtm

PR #709, opened 2023-01-09
fix(mobile): added storage methods without allocation reference

cnlangzi:
### Changes
- added storage.go: exposed file operations without allocation reference between go and Objective-C to avoid crashing on iOS app

### Fixes
-

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


mobilesdk/zbox/cache.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/0chain/gosdk/zboxcore/sdk"
+)
+
+type cachedAllocation struct {
+	Expiration time.Time
+	Allocation *sdk.Allocation
+}
+
+var (
+	cachedAllocations      = make(map[string]*cachedAllocation)
```
peterlimg: Let's use `github.com/hashicorp/golang-lru/v2` for the cache to avoid the cache size increase without limitation. And it should be thread safe as well.

peterlimg (changes requested):

mobilesdk/zbox/cache.go
cnlangzi: fixed

peterlimg (approved):
lgtm

PR #715, opened 2023-01-12
fix(wasm): automatically refresh wallet nonce

cnlangzi:
### Changes
-

### Fixes
- fixed wallet nonce issue. and added automatically refresh wallet nonce in background worker

### Tests

- 0chain: https://github.com/0chain/0chain/pull/1987

### DIFF
--- a/wasmsdk/wallet.go
+++ b/wasmsdk/wallet.go
@@ -1,14 +1,69 @@
+//go:build js && wasm
+// +build js,wasm
+
 package main
 
 import (
+	"time"
+
+	"github.com/0chain/gosdk/core/zcncrypto"
+	"github.com/0chain/gosdk/zboxcore/client"
 	"github.com/0chain/gosdk/zcncore"
 )
 
 func createWallet() (string, error) {
 	return zcncore.CreateWalletOffline()
-
 }
 
 func recoverWallet(mnemonics string) (string, error) {
 	return zcncore.RecoverOfflineWallet(mnemonics)
+}
+
+func setWallet(clientID, publicKey, privateKey string) error {
+	c := client.GetClient()
+	c.ClientID = clientID
+	c.ClientKey = publicKey
+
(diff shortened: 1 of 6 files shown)


wasmsdk/wallet.go
```diff
@@ -1,14 +1,69 @@
...
+	c := client.GetClient()
+	c.ClientID = clientID
+	c.ClientKey = publicKey
+
+	w := &zcncrypto.Wallet{
+		ClientID:  clientID,
+		ClientKey: publicKey,
+		Keys: []zcncrypto.KeyPair{
+			{
+				PrivateKey: privateKey,
```
peterlimg, line 32: Actually it's not safe to have raw private key in memory all the time, theoretically we should only load it when need to sign transaction, and then wipe it immediately. Public key should be enough to get wallet balance. If this setWallet is also being used by other services, then we need to fix the private key in memory issue with a new PR.

peterlimg (changes requested):

wasmsdk/wallet.go
cnlangzi: @peterlimg  in web-apps, wallet is already saved in a global js object. wasm wallet is a copy of the js object. and wasm sdk doesn't know which js object it is in memory. so wasm sdk can't load it on demand.  @ranajahanzaib  do you know other option?

cnlangzi: in web-apps, private key is required to sign almost requests.

peterlimg: okay, we may need to revise the design later, can you create an issue to track it, I'll merge this one first.

peterlimg (approved):
lgtm

wasmsdk/wallet.go
cnlangzi: @peterlimg  raise an issue to discuss it on https://github.com/0chain/gosdk/issues/720

PR #725, opened 2023-01-17
feat(wasm): exposed cancelAllocation and updateAllocation in wasm

cnlangzi:
### Changes
- exposed `getAllocation`, `reloadAllocation`, `cancelAllocation` and `updateAllocation` in wasm

### Fixes
-

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/wasmsdk/blobber.go
+++ b/wasmsdk/blobber.go
@@ -23,7 +23,7 @@ import (
 )
 
 func listObjects(allocationId string, remotePath string) (*sdk.ListResult, error) {
-	alloc, err := getAllocation(allocationId)
+	alloc, err := getAllocation(allocationId, false)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


wasmsdk/blobber.go
```diff
@@ -32,16 +32,16 @@ func listObjects(allocationId string, remotePath string) (*sdk.ListResult, error
 
 }
 
-func createDir(allocationID, remotePath string) error {
-	if len(allocationID) == 0 {
-		return RequiredArg("allocationID")
+func createDir(allocationId, remotePath string) error {
```
peterlimg: We should use `ID` instead of `Id`, so let's use `allocationID`. See https://github.com/golang/go/wiki/CodeReviewComments#initialisms for why we do it.

wasmsdk/cache.go
```diff
@@ -18,16 +19,23 @@ var (
 	cachedAllocations, _ = lru.New[string, *cachedAllocation](100)
 )
 
-func getAllocation(allocationID string) (*sdk.Allocation, error) {
-	it, ok := cachedAllocations.Get(allocationID)
+func getAllocation(allocationId string, refresh bool) (*sdk.Allocation, error) {
```
peterlimg: The refresh may confuse users and may difficult to choose whether they should do refresh or not, or in most of the case, they would simply chose to refresh. So why not remove the param and just do the refresh logic.

peterlimg (changes requested):

wasmsdk/blobber.go
cnlangzi: fixed

wasmsdk/cache.go
cnlangzi: removed `refresh`, and added `reloadAllocation`

peterlimg (approved):
lgtm

PR #746, opened 2023-02-02
fix(zcncore): fixed split keys feature on mobile

cnlangzi:
### Changes
-

### Fixes
- fixed public key issue when wallet key is split.

### Tests

- 0chain: https://github.com/0chain/0chain/pull/2074


zcncore/transactionauth_base.go
```diff
@@ -32,7 +32,7 @@ func newTransactionWithAuth(cb TransactionCallback, txnFee uint64, nonce int64)
 }
 
 func (ta *TransactionWithAuth) getAuthorize() (*transaction.Transaction, error) {
-	ta.t.txn.PublicKey = _config.wallet.Keys[0].PublicKey
+	ta.t.txn.PublicKey = _config.wallet.ClientKey
```
peterlimg, line 35: The `sign()` method below will use `_config.wallet.Key[0].PrivateKey` to sign the transaction, so if it is possible that the `ClientKey` is not the `Keys[0].PublicKey`, then we will use the wrong key to sign the transaction. My question is: is it possible the `_config.wallet.Keys` be in different order that the first key is not the `ClientKey`? If it's always the `ClientKey`, then we actually don't need this changes, otherwise, we'd better make sure the order is guaranteed.

peterlimg (changes requested):

zcncore/transactionauth_base.go
cnlangzi: @peterlimg `_config.wallet.ClientKey` is primary public key.  `_config.wallet.Keys[0].PublicKey` is the public key of `bolt`. otherSign is sign by `_config.wallet.Keys[1].PrivateKey` on `zAuth` desktop.  And the txn should combine 1st sign and 2nd sign. on blockchain it should be verified by primary public key. but `txn.PublicKey` is set with bolt's one. that is why `splitkeys` doesn't work anymore.
BTW: this `transactionauth` is used on `splitkeys` feature only.  so generic transaction will not be affected.

cnlangzi: it always throws `invalid signature` error. because bolt's public key always be sent to blockchain as primary key on old code. that is why we need verify it on blockchain in https://github.com/0chain/0chain/pull/2074

peterlimg: What's the relationship between the primary public key and the bolt's public. All keys in the wallet should be derived from the same seed. Otherwise we need to separate the keys.

cnlangzi: @peterlimg check detail from https://github.com/0chain/gosdk/blob/7140de11080e5b2e45fd6545a0fb662b77fd713f/core/zcncrypto/bls0chain_herumi_test.go#L114

cnlangzi: Keys are spilt from primary private key

cnlangzi: Please also check `SplitKeys` and `SetupAuth` methods for more context

peterlimg: Oh no, who gave the instructions about the implementation of the `SplitKeys`? The `SplitKey` is not deterministic, which means given the same primary key, running it multiple times would give random keys. This will make the wallet unable to recovery from seed, and users can lost their tokens if they think they can recover all keys from seed.

cnlangzi: on bolt, it will ask to recover primary key from could or mnemonics first, and get  new SplitKeys for 2FA. `Recover SplitKeys` seems not be required there. @guruhubb please confirm it.thanks.

peterlimg (approved):
LGTM

service-0chain:
| config  | value |
| tests skipped  | TRUE  |
| system-tests  | master  |
| gosdk  | fix/split_keys_on_mobile  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | fix/split_keys_on_mobile  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | fix/split_keys_on_mobile  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #753, opened 2023-02-06
Enhancement/lp refactor download

### LINKED ISSUE #314: refactor download file
- download fragments by chunks in parallel requests (not `Chunk` on [Protocols](https://github.com/0chain/blobber/wiki/Protocols#what-is-fixedmerkletree)). It is https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests
- download fragments in parallel requests from blobbers 
- resume a download of a file that has been partially transferred.
- still need to support [video-streaming](https://docs.0chain.net/0chain/zbox-cli#video-streaming) . please check more detail with @moldis if there is any question. it is used on local or web video player
- pipe stream support (output bytes to io.Writer instead of file directly)
- support to reconstruct data on local with minimal fragments instead of download all fragments if it is possible . check [here](https://github.com/klauspost/reedsolomon/blob/5593e2b2ddcaeea40d28d7007369105a3bff7e66/reedsolomon.go#L761) for detail

lpoli:
### Changes
Following changes are added:
1. Add reader that complies with io.ReadSeekCloser interface
2. Provide download function for zboxcli to read file into some file; it also serves as an example to how to use reader
 

### Fixes
-

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli: https://github.com/0chain/zboxcli/pull/405
- zwalletcli:
- Other: ...


zboxcore/sdk/reader.go
```diff
@@ -0,0 +1,232 @@
...
+			// otherwise `no shard data` error will occur in erasure reconstruction.
+			numBlocks = endInd - startInd
+		}
+
+		data, err := sd.getBlocksData(startInd, numBlocks)
+		if err != nil {
+			return 0, err
+		}
+
+		offset := sd.offset % int64(effectiveChunkSize)
```
peterlimg, line 151: Not sure what does the mod do here. And seems it could cause data lose. Imagine we request block_1, the offset is 109, and effectiveChunkSize is 100, then we will copy the `data[9:]` in block_1 to `b`. So where's the `data[:9]` going.

peterlimg (changes requested):

zboxcore/sdk/reader.go
lpoli: @peterlimg If client requests with offset 109, then it means client does not want or care about `data[:9]`. So for example, if client already had stored 109 bytes of data then for requesting other data offset would be set to 109. Since data is requested in block level(100 bytes in above example), 109 bytes means client has stored 100 bytes from block_0 and 9 bytes from block_1, so now after getting block_1, it should skip `data[:9]` otherwise data will be corrupted.

peterlimg (approved):
LGTM

PR #755, opened 2023-02-07
Remove sending custom timeout as an option to resty object

### LINKED ISSUE #737: improve `resty` package
as per @peterlimg mentioned on https://github.com/0chain/gosdk/pull/718#discussion_r1089708101

> We'd better use the ctx instead of coming up with a new timeout, the API caller should be able to control the timeout with the ctx param. Anyway, that will require the refactoring of the. resty package to drop the timeout param or make it optional, which could be done in another thread.

aniketsingh03:
### Changes
- This PR removes the custom timeout param that was being sent while creating resty object. The timeout will now be controlled by the API caller/client through the request context. However, a default timeout of 10s (`DefaultRequestTimeout`) is still there for requests which do not have any timeout set.

### Fixes
- Fixes #737 

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/core/resty/option.go
+++ b/core/resty/option.go
@@ -2,7 +2,6 @@ package resty
 
 import (
 	"net/http"
-	"time"
 )
 
 // WithRetry set retry times if request is failure with 5xx status code. retry is ingore if it is less than 1.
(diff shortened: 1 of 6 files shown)


aniketsingh03:
Opened a new PR for this with upstream's branch as compare - #758

PR #759, opened 2023-02-08
Error in Put transaction in chaos environment

Kishan-Dhakan:
### Changes
- does not fail send txn immediately with a non success status code

### Fixes
-

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


peterlimg (approved):
lgtm

zcncore/transaction_base.go
```diff
@@ -361,6 +359,14 @@ func (t *Transaction) submitTxn() {
 				return
 			}
 
+			if res.StatusCode != http.StatusOK {
```
peterlimg, line 364: ah, giving none `StatusOK` means the txn execution failed due to some reason("invalid txn nonce" for example), the txn is actually reached to the miners, and we should return the error to the caller through the `resultC` channel. 

Shouldn't the error returned by`res, err := req.Post()` give the clue about the chaos caused node unreachable info? and I think it should have been handled above.

peterlimg (changes requested):
lgtm

peterlimg:
I'll close this PR as it will hide the true txn failure errors, the chaos caused txn submit failure case should have been handled, but it may slow as we will wait for timeout if all miners were offline. Feel free to re-open if needed.

zcncore/transaction_base.go
Kishan-Dhakan: No, on unreachable nodes status is 400 and req.Post() doesn't return any error.

PR #760, opened 2023-02-09
Enhance the performance of getting a random healthy sharder for transaction API calls

### LINKED ISSUE #756: Improve the performance of getting a random healthy sharder for running transaction queries
The current implementation of getting a random sharder to run transaction queries looks something like this (from `transaction_query_base.go`):
```go
// randOne random one health sharder
func (tq *TransactionQuery) randOne(ctx context.Context) (string, error) {

	randGen := rand.New(rand.NewSource(time.Now().UnixNano()))
	for {

		// reset selected if all sharders were selected
		if len(tq.selected) >= tq.max {
			tq.selected = make(map[string]bool)
		}

		i := randGen.Intn(len(tq.sharders))
		host := tq.sharders[i]

		_, ok := tq.selected[host]

		// it was selected, try next
		if ok {
			continue
		}

		tq.selected[host] = true

		err := tq.checkHealth(ctx, host)

		if err != nil {
			if errors.Is(err, ErrNoOnlineSharders) {
				return "", err
			}

			// it is offline, try next one
			continue
		}

		return host, nil
	}
}
```

This is sequential and will hamper the performance when we go on main net, where there can be lots of sharders. This can be optimized to run in parallel and returning as soon as we receive a response from a healthy sharder.

Let me know if this is something we should focus on and I can write a quick fix for this @dabasov @peterlimg @guruhubb

aniketsingh03:
### Changes
This PR adds concurrency to the part of fetching a random healthy sharder for transaction APIs. Earlier, it was happening on a sequential manner, which would have increased the time to fetch a healthy sharder in proportion to the number of sharders. The concurrent approach will be much faster as the sharders scale.
I have also added some important unit and benchmark tests for the changes made.

### Fixes
Fixes #756 

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


zcncore/transaction_query_base.go
```diff
@@ -143,52 +144,54 @@ func (tq *TransactionQuery) checkHealth(ctx context.Context, host string) error
...
+			select {
+			case <-ctx.Done():
+				finalErrCh <- ctx.Err()
+				return
+			case e := <-errCh:
+				if errors.Is(e, ErrSharderOffline) {
+					tq.RLock()
+					defer tq.RUnlock()
+					if len(tq.offline) >= tq.max {
+						finalErrCh <- ErrNoOnlineSharders
```
dabasov: finalErrCh should be buffered to read all the errors, to unlock the goroutines, or you should force wait for all to finish with wg

dabasov (changes requested):

zcncore/transaction_query_base.go
aniketsingh03: Made some changes here too, removed the extra goroutine and waiting on ctx.Done, cuz it was unnecessary. Pls have a look now.

zcncore/transaction_query_base.go
```diff
@@ -148,49 +148,55 @@ func (tq *TransactionQuery) checkHealth(ctx context.Context, host string) error
...
+func (tq *TransactionQuery) getRandomSharder(ctx context.Context) (string, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	var mu sync.Mutex
+	done := false
+	errCh := make(chan error, len(tq.sharders))
+	successCh := make(chan string)
+
+	for _, sharder := range tq.sharders {
+		go func(sharder string) {
```
peterlimg: This would send requests to all sharders to get one healthy sharder. We should minimize the requests number when possible, so please limit the requests. We can send to 3 sharders concurrently at most, and if all failed, then try next 3 sharders.

peterlimg (changes requested):

zcncore/transaction_query_base.go
aniketsingh03: Okay, but I am wondering if just sending requests to 3 sharders concurrently would be any better than doing it sequentially?

peterlimg: Imagine if the selected 3 sharders were offline, but all other sharders are available. Anyway, what I suggested is to send requests to 3 sharders concurrently, if any one  success, then we return. If all requests failed, then we try next 3 sharders and repeat until find one healthy sharder or return error if no sharders alive.

zcncore/transaction_query_base.go
```diff
@@ -148,49 +148,55 @@ func (tq *TransactionQuery) checkHealth(ctx context.Context, host string) error
...
+				}
+				tq.RUnlock()
+			} else if errors.Is(e, context.DeadlineExceeded) {
+				return "", e
 			}
-
-			// it is offline, try next one
-			continue
+		case s := <-successCh:
+			log.Printf("Found a healthy sharder: %v", s)
```
peterlimg: We'd better avoid logging in sdk when possible.

peterlimg (changes requested):

zcncore/transaction_query_base.go
aniketsingh03: Got it, let me make those changes

aniketsingh03: Done

aniketsingh03:
Please have a look and review again if needed @peterlimg.

zcncore/transaction_query_base.go
```diff
@@ -140,57 +140,81 @@ func (tq *TransactionQuery) checkHealth(ctx context.Context, host string) error
...
-	for {
-
-		// reset selected if all sharders were selected
-		if len(tq.selected) >= tq.max {
-			tq.selected = make(map[string]bool)
+// getRandomSharder returns a random healthy sharder
+func (tq *TransactionQuery) getRandomSharder(ctx context.Context) (string, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	for i := 0; i < len(tq.sharders); i += tq.numShardersToBatch {
```
peterlimg: We will lost the randomness as we always start from the same 3 sharders. We can copy the sharders, shuffle them, and keep all the rest the same but use the shuffled sharders.

peterlimg (changes requested):

zcncore/transaction_query_base.go
aniketsingh03: Thanks for pointing this out. Fixed this in next commit.

peterlimg (changes requested):
@aniketsingh03 we only need to shuffle once at the beginning before the loop start, then all the process logic could be the same as before. Otherwise, if we shuffle it on each batch, we could check the same sharder multiple times and the worst case is we don't see any valid sharder even we have. 
For example, suppose we have sharders [s1, s2, s3, s4, s5, s6], batch size as 3, and only s6 is alive. Basically we will have 2 rounds of batch requests, suppose the first round with shuffled sharders `[s2, s1, s4, s5, s6, s3]`, as the first 3 are all offline, we will go to next round, and if we shuffle it with sharders `[s3, s5, s1, s4, s2, 6]`, then we will return with error that all sharders are not health, which is actually not correct. 

So please just shuffle once at the beginning so that we can iterate all of them to see whether we have health sharders.

aniketsingh03:
The shuffling happens only on the unprocessed sharders, so for the example you mentioned, in the first round, it will shuffle as `[s2, s1, s4, s5, s6, s3]`, run this round on `[s2, s1, s4]` and for the 2nd round it will shuffle the remaining ones and produce `[s3, s5, s6]` and so on so forth until all the sharders have been processed.
But I think this might be an overkill, randomness over randomness doesn't make sense. Let me revert this

peterlimg:
@aniketsingh03 oh, didn't see that it would only shuffle the remaining ones. But let's keep it simple and just shuffle once.

peterlimg (approved):
LGTM

dabasov (approved):
lgtm

PR #787, opened 2023-02-20
Fix: reconciliation problem

YarikRevich:
## Changes
- Adds methods, which are used to retreive from sharders the last processed mint nonce and a list of not processed burn tickets for a certain ethereum address
- Adds WASM methods, which are used to get a list of not processed burn ZCN transaction hashes and not processed burn WZCN transaction hashes
- Adds zcnbridge method, which is used to get a list of not processed burn ZCN transaction hashes
- Adds support of "SubgraphURLAPI" param in bridge.yaml, which is used to establish a connection with "The Graph" GraphQL server


zcncore/http_consensus_objects.go
```diff
@@ -0,0 +1,91 @@
...
+	if c.WinMapConsensus[hash] > c.MaxConsensus {
+		c.MaxConsensus = c.WinMapConsensus[hash]
+		c.WinMap = m
+		c.WinInfo = respBody
+	}
+
+	return nil
+}
+
+func (c *HttpConsensusObjects) buildObject(respBody string) (json.RawMessage, string, error) {
```
dabasov: why can't we resty and already implemented functions for it?

dabasov (changes requested):

zcncore/http_consensus_objects.go
YarikRevich: We cannot use already existing concensus structure, 'cause it was made for the map-like data. I didn't copy the main logic. Unmarshaling process is completely different

YarikRevich: https://github.com/0chain/gosdk/blob/e769e25b952657c78986b07943ab9660213b3f8e/zcncore/transaction_query.go#L316 is not capable to be built for mobile devices

service-0chain:
| config  | value |
| tests skipped  | TRUE  |
| system-tests  | master  |
| gosdk  | feature/reconciliation_problem  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

dabasov (approved):
lgtm

PR #816, opened 2023-03-02
fix: error callback

aaryankotharii:
### Changes
- Error callback was returning local file location, fixed to return remote file path.
- update: `req.filePath` to `req.remotefilepath`


peterlimg (approved):
LGTM

PR #830, opened 2023-03-06
fix: error callback

aaryankotharii:
### Changes
- Error callback was returning local file location, fixed to return remote file path.
- update: `su.fileMeta.Path` to `su.fileMeta.RemotePath`

missed `zboxcore/sdk/chunked_upload.go` in #816


cnlangzi (approved):
LGTM

PR #847, opened 2023-03-14
fix move file error message

Manish-210:
### Changes
-changing moveworker.go to show correct error message.

### Fixes
-https://github.com/0chain/zboxcli/issues/240 
(move file)

### Tests

- blobber:
- 0chain:
- system_test: https://github.com/0chain/system_test/pull/586
- zboxcli: https://github.com/0chain/zboxcli/pull/403
- zwalletcli:
- Other: ...


service-0chain:
| config  | value |
| tests skipped  | ABILITY TO SKIP TESTS IS DISABLED. SKIP_TESTS OPTION IGNORED.  |
| system-tests  | move_file_error_message  |
| gosdk  | fix_incorrect_move_file_error_message  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | ABILITY TO SKIP TESTS IS DISABLED. SKIP_TESTS OPTION IGNORED.  |
| system-tests  | incorrect_move_file_error_message  |
| gosdk  | fix_incorrect_move_file_error_message  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

guruhubb (approved):

zboxcore/sdk/moveworker.go
```diff
@@ -158,15 +162,34 @@ func (req *MoveRequest) ProcessMove() error {
...
 			}
-			objectTreeRefs[blobberIdx] = refEntity
+			l.Logger.Error(err.Error())
+			
 		}(int(pos))
 	}
 
-	wg.Wait()
+	go func() {
+		wg.Wait()
```
dabasov: it will leak if wg.Wait() is blocked forever

dabasov (dismissed):

zboxcore/sdk/moveworker.go
Manish-210: we are using defer wg.Done, So it won't be blocked because for every wg.Add there is wg.Done . Can you please explain it, how it can be blocked forever.

zboxcore/sdk/copyworker.go
```diff
@@ -160,14 +164,33 @@ func (req *CopyRequest) ProcessCopy() error {
...
 				l.Logger.Error(err.Error())
 				return
 			}
 			objectTreeRefs[blobberIdx] = refEntity
 		}(int(pos))
 	}
+	go func() {
+		wg.Wait()
+		close(wgDone)
+	}()
```
peterlimg: We don't need to put the `wg.Wait()` in goroutine, and use `wgDone` channel. We cam simplify it with:
```go
// for loop to copy object the same as above

wg.Wait()
var err error
select {
case err = <-wgErrors:
default:
}

if !req.isConsensusOk() && err != nil {
  return errors.New("copy_failed", fmt.Sprintf("copy failed: %v", err))
}

// keep remaining the same
```

zboxcore/sdk/moveworker.go
```diff
@@ -158,15 +162,38 @@ func (req *MoveRequest) ProcessMove() error {
...
+				default:
 			}
-			objectTreeRefs[blobberIdx] = refEntity
+			l.Logger.Error(err.Error())
+			
 		}(int(pos))
 	}
 
-	wg.Wait()
+	go func() {
```
peterlimg: Same as above

zboxcore/sdk/moveworker.go
```diff
@@ -158,15 +162,38 @@ func (req *MoveRequest) ProcessMove() error {
 		go func(blobberIdx int) {
 			defer wg.Done()
 			refEntity, err := req.moveBlobberObject(req.blobbers[blobberIdx], blobberIdx)
-			if err != nil {
-				l.Logger.Error(err.Error())
-				return
+
+			if err == nil {
```
peterlimg: Let's do `if err != nil`

peterlimg (changes requested):

zboxcore/sdk/copyworker.go
Manish-210: @peterlimg 
If we don't use wgDone channel then
https://github.com/0chain/gosdk/actions/runs/4460231710/jobs/7833235220?pr=847 
- **by using wg.wait() in goroutine:** the go routines will all be completed but no error will be received from the wgErrors channel because imo default will be executed before even the error is pushed in channel.
- **by without using wg.wait() in goroutine:** it will create panic

Is it possible to do so in some other way?

peterlimg: @Manish-210 wg.wait() is suppose to run outside of the goroutines. The persuade code would be:

```go
wg := sync.WaitGroup{}
for i := 0; i < n; i++ {
  wg.Add(1)
  go func() {
      defer wg.Done()
      if err := doSomthing(); err != nil {
        select {
        case wgErrors <- err:
        default:
        }
        return
      }
}()
}

wg.Wait() 
// all goroutines have finished
select {
case err := wgErrors:
default:  // it would not got to default case if any error happens in the goroutine above, it is because `wg.Wait()` guarantees that all goroutines have finished, which means the error must have been pushed to channel or ignored if the channel is full.
}
```
Do you mean we will panic if we do it in this way? but how?

Manish-210: Thanks @peterlimg for the suggestion
https://go.dev/play/p/PLe31txueLn
I tried to implement the same approach. But it will create a deadlock cause we are listening after the goroutines have been completed wgErrors channel.

peterlimg: Try again, I changed a little bit, we need to use `select ... default` case when pushing err to `wgErrors` channel in case the channel is full.

peterlimg (approved):
LGTM

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | incorrect_move_file_error_message  |
| gosdk  | fix_incorrect_move_file_error_message  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | incorrect_move_file_error_message  |
| gosdk  | fix_incorrect_move_file_error_message  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

cnlangzi (changes requested):
`wgErrors`'s buffer is 1, and `wg.Wait` has to wait for all requests are done.  It makes deadlock if any error is thrown, and send to `wgErrors`. @peterlimg  could you check it again?
FYI: it is have been refactored in `rename` command.

zboxcore/sdk/copyworker.go
```diff
@@ -160,15 +162,30 @@ func (req *CopyRequest) ProcessCopy() error {
...
 			}
 			objectTreeRefs[blobberIdx] = refEntity
 		}(int(pos))
 	}
-
 	wg.Wait()
 
+	var recentErr *error
+	select {
+		case err := <-wgErrors:
```
cnlangzi: line 153 + line 166 + line 175 + line 179 might make deadlock. @Manish-210  could you add an unit test to verify it?

cnlangzi (dismissed):

zboxcore/sdk/copyworker.go
Manish-210: @cnlangzi 
The unit test here will have 2 blobbers giving correct responses from the 4 blobbers and it passes this test.

https://github.com/0chain/gosdk/blob/fix_incorrect_move_file_error_message/zboxcore/sdk/copyworker_test.go#L420

Is there any other unit test to be added?

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | incorrect_move_file_error_message  |
| gosdk  | fix_incorrect_move_file_error_message  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

zboxcore/sdk/copyworker.go
peterlimg: @cnlangzi it would not create deadlock, the `select...default` will unblock it.

peterlimg:
> `wgErrors`'s buffer is 1, and `wg.Wait` has to wait for all requests are done. It makes deadlock if any error is thrown, and send to `wgErrors`. @peterlimg could you check it again? FYI: it is have been refactored in `rename` command.

@cnlangzi only the first error will be pushed to the `wgErrors` channel, the following errors will be ignored by the `default` case as the buffer is full.

zboxcore/sdk/copyworker.go
cnlangzi: @peterlimg thanks. I get the idea about `select ... default` now. but it smells as a tricky workaround. could we implement it with more clean and readable style with buffer channel ? eg
```
func main() {

	wgErrors := make(chan error, 10)
	wg := &sync.WaitGroup{}

	for i := 0; i < 10; i++ {

		wg.Add(1)
		go func(blobberIdx int) {
			defer wg.Done()
                        err := ....
			if err != nil {

				wgErrors <- err
			}
			return
		}(i)
	}
	wg.Wait()

	if len(wgErrors) > 0 {
		err := <-wgErrors
		fmt.Println(err)
	}
}
```

peterlimg: yes, ideally we should use buffered channel, the following changes that @Manish-210 has done should be fine as well.

PR #886, opened 2023-04-03
Fix lockBlobber sleep

### LINKED ISSUE #879: Need a better way to implement lockBlobber function in writemarker_mutex.go
In the current implementation of [lockBlobber](https://github.com/0chain/gosdk/blob/eb0f6b0a69d87c034a1948c5bafd0aa8fab64e81/zboxcore/sdk/writemarker_mutex.go#L242), if the blobbers are locked, then the current request goes to sleep for `2*timeout` time, so when it again make a request, then blobbers are returning a timeout error. 
This would create a problem if we wanted to upload/update multiple files parallelly through the `sync` command. 
We need to correct this before closing [#100](https://github.com/0chain/zboxcli/issues/100)

https://github.com/0chain/gosdk/blob/eb0f6b0a69d87c034a1948c5bafd0aa8fab64e81/zboxcore/sdk/writemarker_mutex.go#L316

Roshan-Mehta:
### Changes
- Decreased sleep time if the blobber is already locked.
- Try indefinitely if the lock status is pending

### Fixes
- Fix https://github.com/0chain/gosdk/issues/879

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


zboxcore/sdk/writemarker_mutex.go
```diff
@@ -313,7 +313,7 @@ func (wmMu *WriteMarkerMutex) lockBlobber(
 				if wmLockRes.Status == WMLockStatusPending {
 					logger.Logger.Info("Lock pending for blobber ",
 						b.Baseurl, "with connection id: ", connID, " Retrying again")
-					time.Sleep(timeOut * 2)
+					time.Sleep(5 * time.Second)  // sleep for 5 seconds
```
lpoli: Please use constant variable to declare this duration.
Also is it only the place that requires this change?

lpoli (changes requested):

zboxcore/sdk/writemarker_mutex.go
Roshan-Mehta: @lpoli  
`Also is it only the place that requires this change?`
The other potential place is [this](https://github.com/0chain/gosdk/blob/a902336c8a93df9eee5daac96244ebbdb849f0ad/zboxcore/sdk/writemarker_mutex.go#L326), where `resp.StatusCode == http.StatusAccepted`, but I couldn't reproduce this response from blobber, so can't comment much. Shall I change this also?

zboxcore/sdk/writemarker_mutex.go
```diff
@@ -363,6 +363,7 @@ func (wmMu *WriteMarkerMutex) lockBlobber(
 		if !shouldContinue {
 			break
 		}
+		WMLockWaitTime *= 2 //In each round we increase the wait time by factor 2.
```
Roshan-Mehta: Instead of fix wait time, we are increasing wait time
by factor 2 and starting with low wait time. When there are multiple
goroutines (say n)  trying to lock Blobber, then we no longer need to wait
O(n * wait_time) time.

lpoli: Lets not make WMLockWaitTime a variable instead make it constant.
Waittime is increasing at some condition and never decreasing.
For simplicity lets put:
`const WMLockWaitTime = time.Second * 5`
To wait simply multiply it by some factor but never update the value

lpoli (changes requested):

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | fix/lock-blobber-sleep  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0dns  | staging  |

zboxcore/sdk/writemarker_mutex.go
```diff
@@ -242,7 +243,7 @@ func (wmMu *WriteMarkerMutex) Lock(
 func (wmMu *WriteMarkerMutex) lockBlobber(
 	ctx context.Context, mask *zboxutil.Uint128, maskMu *sync.Mutex,
 	consensus *Consensus, b *blockchain.StorageNode, pos uint64, connID string,
-	requestTime string, timeOut time.Duration, wg *sync.WaitGroup) {
+	requestTime *string, timeOut time.Duration, wg *sync.WaitGroup) {
```
lpoli: Why do we need to send requestTime pointer?

Roshan-Mehta: Because we are refreshing the lock after every 20 seconds in Lock() function. So, Lock() function needs to know the requestTime.

lpoli: The request time was already being updated in line 224 so we do not need to update it in line 278.
And Lock() function does not need to know about request time, though irrelevant to the usecase

Roshan-Mehta: Yes, correct. I think I need to send the pointer of "requestTime" from the Lock() function and not the pointer of "rT".

Roshan-Mehta: We need to send requestTime as a pointer. Let's say From Lock() function, there is a call to lockBlobber at some time (say at the 0th second), but the lock is acquired at the 15th second, so the lock should refresh at the 35th second (after the 20th second of lock acquired). So, my point is Lock() function should know when the lock is actually acquired by LockBlobber().

Roshan-Mehta: Anyway, we are removing request time (https://github.com/0chain/gosdk/pull/926), so it doesn't matter now.

dabasov:
we won't use request marker in lock request

zboxcore/sdk/writemarker_mutex.go
```diff
@@ -308,21 +309,15 @@ func (wmMu *WriteMarkerMutex) lockBlobber(
...
+					time.Sleep(WMLockWaitTime)
 					shouldContinue = true
+					retry--
 					return
 				}
 				err = fmt.Errorf("Lock acquiring failed")
 				return
 			}
 
-			if resp.StatusCode == http.StatusAccepted { // accepted but pending
```
peterlimg, line 319: We should keep the code for processing the `StatusAccepted`, we need to wait and retry, otherwise, we will try to read body from the response and return error, which is not what expected.

peterlimg (changes requested):

zboxcore/sdk/writemarker_mutex.go
Roshan-Mehta: @peterlimg Does the blobber return StatusAccepted ever? I saw the blobber code and couldn't find it. Can you point to the file?

peterlimg: @Roshan-Mehta yes, you're right, the blobber does not return StatusAccepted code, don't know why it checks it here before. So should be safe to remove it.

peterlimg (approved):
LGTM

PR #915, opened 2023-04-24
Fix/EOF

### LINKED ISSUE #889: EOF when downloading file from blobber
File download is returning with EOF error.

**EOF** is basically responded when client uses a http connection that is closed by the server.
Its not easy to replicate this error and fix it but having connection ttl should have been fine.

Network details : 

2 Blobbers and 2 Validators 

Blobber1 : 0 read_price | 0.1 write_price
Blobber2 : 0.01 read_price | 0.1 write_price

Allocation Details : 

1 data 1 parity
1 GB size 

File Size : 0.1GB 

Both blobbers and validators are staked 1 tokens each.

lpoli:
### Changes
-

### Fixes
-  Tries to fix [EOF issue](https://github.com/0chain/gosdk/issues/889). The cause of EOF is unknown but it basically seems to be related to closed connection issue or issue with readLoop in net/http package. 
References:

1. https://github.com/googleapis/google-cloud-go/issues/7090
2. https://github.com/golang/go/issues/53472
3. https://stackoverflow.com/questions/17714494/golang-http-request-results-in-eof-errors-when-making-multiple-requests-successi

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/zboxcore/zboxutil/http.go
+++ b/zboxcore/zboxutil/http.go
@@ -741,7 +741,21 @@ func MakeSCRestAPICall(scAddress string, relativePath string, params map[string]
 func HttpDo(ctx context.Context, cncl context.CancelFunc, req *http.Request, f func(*http.Response, error) error) error {
 	// Run the HTTP request in a goroutine and pass the response to f.
 	c := make(chan error, 1)
-	go func() { c <- f(Client.Do(req.WithContext(ctx))) }()
+	go func() {
+		var err error
+		// indefinitely try if io.EOF error occurs. As per some research over google
+		// it occurs when client http tries to send byte stream in connection that is
+		// closed by the server
+		for {
+			err = f(Client.Do(req.WithContext(ctx)))
+			if errors.Is(err, io.EOF) {
+				continue
+			}
+			break
+		}
+		c <- err
+	}()
+
 	// TODO: Check cncl context required in any case
 	// defer cncl()
 	select {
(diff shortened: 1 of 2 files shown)


zboxcore/zboxutil/http.go
```diff
@@ -741,7 +741,21 @@ func MakeSCRestAPICall(scAddress string, relativePath string, params map[string]
 func HttpDo(ctx context.Context, cncl context.CancelFunc, req *http.Request, f func(*http.Response, error) error) error {
 	// Run the HTTP request in a goroutine and pass the response to f.
 	c := make(chan error, 1)
-	go func() { c <- f(Client.Do(req.WithContext(ctx))) }()
+	go func() {
+		var err error
+		// indefinitely try if io.EOF error occurs. As per some research over google
+		// it occurs when client http tries to send byte stream in connection that is
+		// closed by the server
+		for {
```
dabasov, line 749: use context to break this loop and terminate goroutine properly

```
for {
    select {
        case <-ctx.Done():
             return
        default:
             //your code
    }
}
```

dabasov (changes requested):

zboxcore/zboxutil/http.go
lpoli: Context is already in use inside `client.Do` function. Only for EOF error it will continue. If it frequently shows up EOF error then it will anyway break when context timeout occurs

dabasov: ok, it should be enough

dabasov (approved):
lgtm

PR #949, opened 2023-05-04
Fix: nil reference array as a response for burn events retrieval request

### LINKED ISSUE #948: Fix: nil reference array as a response for burn events retrieval request

YarikRevich:


dabasov (approved):
lgtm

PR #966, opened 2023-05-15
return error if allocation is cancelled

boddumanohar:
### Changes
-

### Fixes
-

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


peterlimg (changes requested):
It can be either `Finalized` or `Canceled`. We should still return the allocation even it's finalized or canceled, but no need to `initialize` it if finalized/canceled. It's user's responsibility to determine whether to continue update the allocation or not. Also, all the actions agains the allocation should check if the allocation is finalized/canceled before sending transaction to reduce unnecessary txn fee.

boddumanohar:
cancelling this PR as  It's user's responsibility to determine whether to use the allocation or not

PR #998, opened 2023-05-29
MultiOp fix conflicts

din-mukhammed:
### Changes
DoMultiOperation will process and commit operations in between if conflicts are found

### Fixes
-

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


dabasov (approved):
lgtm

zboxcore/sdk/allocation.go
```diff
@@ -667,63 +667,99 @@ func (a *Allocation) DoMultiOperation(operations []OperationRequest) error {
...
 
-		case constants.FileOperationRename:
-			operation = NewRenameOperation(op.RemotePath, op.DestName, mo.operationMask, mo.maskMU, mo.consensusThresh, mo.fullconsensus, mo.ctx)
+	for i := 0; i < len(operations); {
+		// resetting multi operation and previous paths for every batch
+		mo = MultiOperation{
+			allocationObj: a,
+			operationMask: zboxutil.NewUint128(1).Lsh(uint64(len(a.Blobbers))).Sub64(1),
+			maskMU:        &sync.Mutex{},
+			ctx:           a.ctx,
```
Roshan-Mehta: ctxCncl is missing. Panic issue in system test might be because of this.

peterlimg (approved):
lgtm

Hitenjain14:
If we want to upload 10 files in a dir ("/a"), it will mark "/a" as true and will have a conflict on every operation and thus will do commit operation 10 times or am I missing something?
cc: @dabasov @Roshan-Mehta

din-mukhammed:
> If we want to upload 10 files in a dir ("/a"), it will mark "/a" as true and will have a conflict on every operation and thus will do commit operation 10 times or am I missing something? cc: @dabasov @Roshan-Mehta

There should be no conflict in such case, fixed it. How about operations on exact same path? Previously there was this check that didn't allow such cases, should we keep this check?
https://github.com/0chain/gosdk/blob/a5173934488d354ee4d009f909958abb2809e843/zboxcore/sdk/allocation.go#L712-L714

Hitenjain14:
> > If we want to upload 10 files in a dir ("/a"), it will mark "/a" as true and will have a conflict on every operation and thus will do commit operation 10 times or am I missing something? cc: @dabasov @Roshan-Mehta
> 
> There should be no conflict in such case, fixed it. How about operations on exact same path? Previously there was this check that didn't allow such cases, should we keep this check?
> 
> https://github.com/0chain/gosdk/blob/a5173934488d354ee4d009f909958abb2809e843/zboxcore/sdk/allocation.go#L712-L714

Marking only the remotePath won't work, we need all the paths from the rootDir and mark them according to the operation.

PR #908, opened 2023-04-20
expose wasm: updateAllocationWithRepair

boddumanohar:
### Changes
- exposes a function to updateAllocationWithRepair
- add a new gosdk function to get Filestats from a single blobber

### Fixes
- 

### Tests
Tasks to complete before merging PR:
- [ ]  Ensure system tests are passing. If not [Run them manually](https://github.com/0chain/gosdk/actions/workflows/system_tests.yml) to check for any regressions :clipboard:
- [ ]  Do any new system tests need added to test this change? do any existing system tests need updated? If so create a PR at [0chain/system_test](https://github.com/0chain/system_test)
- [ ]  Merge your system tests PR to master AFTER merging this PR

### Associated PRs (Link as appropriate):
- blobber:
- 0chain:
- system_test: https://github.com/0chain/system_test/pull/633
- zboxcli: https://github.com/0chain/zboxcli/pull/401
- zwalletcli:
- Other: ...

PR #1003, opened 2023-05-30
fix download after updateallocation

boddumanohar:
### Changes

problem: after I try to update an allocation, if I try to download a file, I get error. 

this was working earlier because we are sending path hash. 

But in the optimize download PR we changed it and we are passing remotefilepath only.

Ideally the blobber should accepts should either of them. But because of a bug on the blobber side, passing pathhash only works.

### Fixes
-

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/zboxcore/sdk/downloadworker.go
+++ b/zboxcore/sdk/downloadworker.go
@@ -603,9 +603,10 @@ func (req *DownloadRequest) attemptSubmitReadMarker(blobber *blockchain.StorageN
 		return fmt.Errorf("error creating download request: %w", err)
 	}
 
+	pathHash := fileref.GetReferenceLookup(req.allocationID, req.remotefilepath)
 	header := &DownloadRequestHeader{
 		Path:         req.remotefilepath,
-		PathHash:     req.remotefilepathhash,
+		PathHash:     pathHash,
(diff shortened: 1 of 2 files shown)


zboxcore/sdk/downloadworker.go
```diff
@@ -802,6 +803,42 @@ type blobberFile struct {
...
+	listReq.consensusThresh = 1
+	listReq.ctx = ctx
+	listReq.remotefilepath = remotePath
+
+	listReq.wg = &sync.WaitGroup{}
+	listReq.wg.Add(1)
+	rspCh := make(chan *fileMetaResponse, 1)
+	go listReq.getFileMetaInfoFromBlobber(listReq.blobbers[0], 0, rspCh)
+	listReq.wg.Wait()
+	resp := <-rspCh
```
peterlimg, line 838: we need to check err in the resp, or return `resp.fileref, resp.err`.

peterlimg (changes requested):

boddumanohar:
already fixed in https://github.com/0chain/gosdk/pull/908

PR #1056, opened 2023-06-20
added caching instead of the network call to estimate the transaction fees

### LINKED ISSUE #1007: use fees_table from sharders to estimate transaction fee
At the moment, to estimate transaction fee we make a call to miners
https://github.com/0chain/gosdk/blob/e00f0fb7e2c44828bffce61be6163e85675d2c45/core/transaction/entity.go#L296-L297

But the transaction fees are always the same a similar endpoint is available on sharders
https://dev1.zus.network/sharder01/v1/fees_table

there is also a function in gosdk: https://github.com/0chain/gosdk/blob/a959cd93adfe82e6c9fa13ce2a47b29a7f3dd7db/core/transaction/entity.go#L358

We can cache the results because these values are almost never updated

SrLeet03:
### Changes
- Implement caching to avoid network calls for retrieving values.

### Fixes
- None  

### Details
- Create a cache object with a capacity of 100 to store the retrieved values.
- Check if the value exists in the cache before making a network call.
- If the value is found in the cache, retrieve it from the cache and return it.
- If the value is not found in the cache, make a network call to fetch it and store it in the cache for future use.
- Update the relevant code sections to utilise the caching mechanism.
                  
This PR improves performance by reducing network calls and improving response times for subsequent requests.

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


core/transaction/entity.go
```diff
@@ -441,3 +487,56 @@ func GetFeesTable(miners []string, reqPercent ...float32) (map[string]map[string
 
 	return nil, errors.New("failed to get fees table", strings.Join(errs, ","))
 }
+
+func GetFeesFromCache() {
```
dabasov: what is it for?

sdkversion
dabasov, line 1: remove this

sdkversion.go
dabasov, line 1: what is it for?

dabasov (changes requested):

core/transaction/entity.go
SrLeet03: sure, will remove that

SrLeet03:
Also, @dabasov, could you help point out where I should put this initLRUCache function?

sdkversion.go
SrLeet03: At the time of local setup of gosdk @boddumanohar asked me the same, sorry forgot to remove it before pushing .

dabasov (approved):
lgtm

cnlangzi (changes requested):
It might not work properly in zbox/zwallet and wasm.
- re zbox/zwallet, one command is one instance. it can't read cache from previous instance
- re wasm, one session is one instance. it can't read cache from other session/instance. 

I think it is better to add persistent storage support
- re zbox/zwallet, it should be cached to ~/.zcn/fees.cache
- re wasm, it should be cached to localStorage

core/transaction/entity.go
```diff
@@ -269,15 +271,34 @@ func sendTransactionToURL(url string, txn *Transaction, wg *sync.WaitGroup) ([]b
...
 }
 
+type cachedObject struct {
+	Expiration time.Duration
+	Value      interface{}
+}
+
+var cache *lru.Cache
+
+func initLRUCache() {
```
dabasov: use once and package level function `init()`, check how it is done for nonceCache

PR #1105, opened 2023-07-11
Optimize repair

### LINKED ISSUE #1053: repair allocation optimisation
currently during allocation repair, we download the entire file, divide into shards as per the updated allocation. This method has 2 problems: 
1. This is solution is not scalable with respect to the size of file.
2. This solution is not applicable in browsers. 

An optimisation that can be done is to do repair on the fly by chunks. You could download a few chunks of data and shard it and upload it to the new blobber. 

this downloaded chunk of data needs to be stored in ring buffer so that we don't take up any space on the local file system. This technique is very much needed to implement repair in browser.

Hitenjain14:
### Changes
-

### Fixes
-

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


dabasov (approved):
lgtm

zboxcore/sdk/repairCallback.go
```diff
@@ -21,10 +21,12 @@ func (s *StatusBar) InProgress(allocationId, filePath string, op int, completedB
...
 	s.success = true
 	l.Logger.Info("Repair for file completed. File = ", filePath)
+	l.Logger.Info("Repair for file completed. File = ", filePath)
 }
 
 func (s *StatusBar) Error(allocationID string, filePath string, op int, err error) {
 	s.success = false
+	s.err = err
 	defer s.wg.Done()
 	defer mutUnlock(s.allocID)
```
sunilmhta, line 31: Are we sure here that mutUnlock won't be called on an allocation twice or it can't be called before calling lock?

PR #1114, opened 2023-07-14
Updating the validator node which is being passed to 0chain in update settings call

sunilmhta:
### Changes
-

### Fixes
-

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/zboxcore/sdk/sdk.go
+++ b/zboxcore/sdk/sdk.go
@@ -621,7 +621,7 @@ type Validator struct {
 }
 
 type UpdateValidator struct {
-	ID                       common.Key        `json:"id"`
+	ID                       common.Key        `json:"validator_id"`
 	BaseURL                  *string           `json:"url,omitempty"`
 	DelegateWallet           *string           `json:"delegate_wallet,omitempty"`
 	MinStake                 *common.Balance   `json:"min_stake,omitempty"`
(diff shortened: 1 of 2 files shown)


peterlimg (approved):
LGTM

zboxcore/sdk/sdk.go
```diff
@@ -636,18 +636,29 @@ type UpdateValidator struct {
...
+		NumDelegates:   v.NumDelegates,
+		ServiceCharge:  v.ServiceCharge,
+	}
+
+	if v.DelegateWallet != nil ||
+		v.MinStake != nil ||
+		v.MaxStake != nil ||
+		v.NumDelegates != nil ||
+		v.ServiceCharge != nil {
+		blockValidator.StakePoolSettings = sp
```
peterlimg, line 658: So we will only update the stake setting when all fields are not nil, right?

peterlimg, line 658: so stake pool settings will be updated only when all fields are set?

sunilmhta: No, it will be set when atleast one of the field is set (hence the condition `or` on each field being `non-nil`), otherwise it will be nil.

PR #1123, opened 2023-07-18
Fix panic in repaircallback

### LINKED ISSUE #1122: Panic in zboxcore/sdk/repairCallback.go
### Context
While load testing (20 allocations with 10k file operations) we observed a panic incidence from gosdk. The error was:  `"log":"fatal error: sync: unlock of unlocked mutex\n"`
The panic happened in repairCallback.go in gosdk in the below method:
```go
func (s *StatusBar) Error(allocationID string, filePath string, op int, err error) {
	s.success = false
	defer s.wg.Done()
	defer mutUnlock(s.allocID)

	var errDetail interface{} = "Unknown Error"
	if err != nil {
		errDetail = err.Error()
	}

	l.Logger.Error("Error in status callback. Error = ", errDetail)
}
```
I suspect this is because of `defer mutUnlock(s.allocID)` line.  We are trying to unlock a mutex on which the lock has not been taken. 

Seems to be a bug here.  

### Fix
We can probably fix this by using some additional indicator (bool?)  to first check if it's already locked or not and then only call unlock?

sunilmhta:
### Changes
-

### Fixes
-

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/zboxcore/sdk/repairCallback.go
+++ b/zboxcore/sdk/repairCallback.go
@@ -26,8 +26,12 @@ func (s *StatusBar) Completed(allocationId, filePath string, filename string, mi
 func (s *StatusBar) Error(allocationID string, filePath string, op int, err error) {
 	s.success = false
 	defer s.wg.Done()
+	if s.unlocked {
+		return
+	}
 	defer mutUnlock(s.allocID)
 
+	s.unlocked = true
 	var errDetail interface{} = "Unknown Error"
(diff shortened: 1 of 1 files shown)


sunilmhta:
should be fixed by https://github.com/0chain/gosdk/pull/1105

PR #1132, opened 2023-07-20
Fix repair cb

Hitenjain14:
### Changes
-

### Fixes
-

### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/zboxcore/sdk/repairCallback.go
+++ b/zboxcore/sdk/repairCallback.go
@@ -47,14 +47,14 @@ type StatusBar struct {
 }
 
 func NewRepairBar(allocID string) *StatusBar {
+	mapLock.Lock()
 	if _, ok := mutMap[allocID]; !ok {
-		mapLock.Lock()
 		mutMap[allocID] = &sync.Mutex{}
-		mapLock.Unlock()
 	}
 	if !mutMap[allocID].TryLock() {
 		return nil
 	}
+	mapLock.Unlock()
 	return &StatusBar{
 		wg:      &sync.WaitGroup{},
 		allocID: allocID,
--- a/zboxcore/sdk/repairCallback.go
+++ b/zboxcore/sdk/repairCallback.go
@@ -48,13 +48,13 @@ type StatusBar struct {
 
 func NewRepairBar(allocID string) *StatusBar {
 	mapLock.Lock()
+	defer mapLock.Unlock()
 	if _, ok := mutMap[allocID]; !ok {
 		mutMap[allocID] = &sync.Mutex{}
 	}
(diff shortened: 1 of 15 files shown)


zboxcore/sdk/repairCallback.go
sunilmhta, line 1: In line 50, where we read from the map: `if _, ok := mutMap[allocID]; !`, I think this should also be within a mutex. Also, on the map, we could probably use `ReaderWriterMutex` instead of simple `mutex`?

sunilmhta: Or we could use Map from `sync` package?

zboxcore/sdk/repairCallback.go
```diff
@@ -25,39 +25,36 @@ func (s *StatusBar) Completed(allocationId, filePath string, filename string, mi
...
+	mapLock.Lock()
 	if _, ok := mutMap[allocID]; !ok {
-		mapLock.Lock()
 		mutMap[allocID] = &sync.Mutex{}
-		mapLock.Unlock()
 	}
 	if !mutMap[allocID].TryLock() {
 		return nil
 	}
+	mapLock.Unlock()
```
sunilmhta: This unlock won't be called if the code execution goes in line 55. Probably it's better to use `defer mapLock.Unlock()` after line 50?

sunilmhta: I think even now this might be a problem. Should we instead use ReadWriterLock for read and write separately?

sunilmhta: We could use a read write lock somethnig like this:
```
// Use a read lock to check if the key exists in the map.
mapLock.RLock()
if _, ok := mutMap[allocID]; !ok {
	mapLock.RUnlock()

	// If the key does not exist, use a write lock to update the map.
	mapLock.Lock()
	mutMap[allocID] = &sync.Mutex{}
	mapLock.Unlock()
} else {
	mapLock.RUnlock()
}

// Now, use a write lock to TryLock the mutex for the allocationID.
mapLock.Lock()
defer mapLock.Unlock()

if !mutMap[allocID].TryLock() {
	return nil
}
```

Hitenjain14: We can defer it and it should be fine, write will take no time so should be alright.

sunilmhta: We would also need to check if this is even a good thing to do and this does not cause starvation to other threads :( 
I think using read lock and then the write lock might NOT be a good idea.

PR #1248, opened 2023-10-06
Feat/exposed cli methods in winsdk

cnlangzi:
### Fixes
- improved InitSDK methods in winsdk
- added `GetFreeStorage` in zboxapi package
- added `CreateFreeStorage` in winsdk
- added `CreateWallet`, `RecoverWallet` in winsdk
- added `List` and `ListAll` in winsdk
- added `GetFileContentType` and `SetFFmpegPath`
- added `ThumbnailHash` and `ThumbnailSize` in ListResult
- added `GetFreeMarker` in WinSDK.
- added `CreateAuthTicket` and `CreateAuthTicket` for sharning
- added `CreateShareInfo`, `GetSharedByMe`, `GetSharedToMe`, `GetSharedByPublic` and `DeleteSharedInfo` in zboxapi
- exported `CreateShareInfo`, `GetSharedByMe`, `GetSharedToMe`, `GetSharedByPublic` and `DeleteSharedInfo` in winsdk
### Tests

- blobber:
- 0chain:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


zboxapi/sdk.go
```diff
@@ -13,8 +13,34 @@ import (
 	"github.com/0chain/gosdk/core/encryption"
 	"github.com/0chain/gosdk/core/resty"
 	"github.com/0chain/gosdk/core/sys"
+	"github.com/0chain/gosdk/core/zcncrypto"
 )
 
+func init() {
+	if sys.Sign == nil {
```
dabasov: we have init function in `zboxcore/client/entity.go` which initializes the same value like this, it is possible that it can be reinitialized since the order of init functions is not defined, please use once() to allow only single initialization

```
func init() {
	client = &Client{
		Wallet: &zcncrypto.Wallet{},
	}

	sys.Sign = signHash
	// initialize SignFunc as default implementation
	Sign = func(hash string) (string, error) {
		return sys.Sign(hash, client.SignatureScheme, GetClientSysKeys())
	}

	sys.Verify = VerifySignature
	sys.VerifyWith = VerifySignatureWith
}
```

zboxapi/sdk.go
```diff
@@ -209,3 +235,36 @@ func (c *Client) RefreshJwtToken(ctx context.Context, phoneNumber string, token
 
 	return result.Token, nil
 }
+
+func (c *Client) GetFreeStorage(ctx context.Context, phoneNumber, token string) (string, error) {
```
dabasov: maybe we can add in the name, that this particular func can be used only with jwtToken, something like `GetFreeStorageWithJWT`

dabasov (dismissed):

zboxapi/sdk.go
cnlangzi: in zboxapi, zbox package is not imported. let me try to improve it.

zboxapi/sdk.go
cnlangzi: token is required in almost apis of 0box. And a jwt token and a firebase token both works. so `GetFreeStorageWithJWT` doesn't make sense.  would you like to rename them with `With` suffix?
eg `GetFreeStorageWith`, `RefreshJwtTokenWith`?  
I would suggest to leave them as they are now?
FYI: `GetFreeStorage` is update. It works with jwt token and firebase token now.

zboxapi/sdk.go
cnlangzi: used local `signHash` instead of `sys.SignFunc`

peterlimg (approved):
lgtm

service-0chain:
| config  | value |
| tests skipped  | false  |
| system-tests  |   |
| gosdk  | feat/zwallet_in_winsdk  |
| zboxcli  |   |
| zwalletcli  |   |
| 0chain  |   |
| blobber  |   |
| authorizer  |   |
| 0box  |   |
| 0dns  |   |

PR #1269, opened 2023-11-01
Revert "enabled gosdk update for sprint branches"

shahnawaz-creator:
Reverts 0chain/gosdk#1268


shahnawaz-creator:
It needs to work by raising pr to sprint branches

PR #1371, opened 2024-01-19
fix: downloadBlocks wasm error

ashuvssut:
# changes
- `downloadBlocks` WASM function was not present in the final WASM build and errors out the following:

	- image
	- upon calling `downloadBlocks`, always `undefined` was returned


ashuvssut:
@Hitenjain14 i have used the current downloadBlocks wasm function and it works as expected in webapps

# Demo
https://github.com/0chain/gosdk/assets/60546840/80c2915d-bd3f-4041-a324-155612581e1d

Hitenjain14 (approved):
lgtm

ashuvssut:
@dabasov @peterlimg Please take a moment to review the changes

wasmsdk/blobber.go
```diff
@@ -839,11 +839,19 @@ func upload(allocationID, remotePath string, fileBytes, thumbnailBytes []byte, w
...
 
 // download download file blocks
-func downloadBlocks(alloc *sdk.Allocation, remotePath, authTicket, lookupHash string, startBlock, endBlock int64) ([]byte, error) {
+func downloadBlocks(allocId string, remotePath, authTicket, lookupHash string, startBlock, endBlock int64) ([]byte, error) {
+
 	if len(remotePath) == 0 && len(authTicket) == 0 {
 		return nil, RequiredArg("remotePath/authTicket")
 	}
 
+	alloc, err := getAllocation(allocId)
```
dabasov, line 848: why do we need to re-request the allocation if we have an object already`?

ashuvssut: > we have an object already

The wasm logged out `downloadBlocks binder not implemented` because `downloadBlocks` wasm function did’t work when it took `*sdk.Allocation` as parameter

so I passed the `allocId` string instead of `*sdk.Allocation` as parameter

Hitenjain14: We cannot pass native go objects from js, so we need to pass allocationID, we have a LRU for allocations so no call will be made to sharders if alloction is already requested before.

ashuvssut:
@dabasov, I've completed the MP4 streaming implementation on the frontend.

Now, the only pending task is to get this GoSDK PR merged because my frontend changes depend on the modifications proposed in this PR.

dabasov (approved):
lgtm

dabasov (approved):

PR #1443, opened 2024-03-27
Feature: add gas amount estimation feature to zcnbridge

### LINKED ISSUE #1442: Feature: add gas amount estimation feature to zcnbridge

YarikRevich:


go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/0chain/gosdk
 
-go 1.21
-
-toolchain go1.21.5
+go 1.18
```
dabasov: what is the reason to downgrade the version?

YarikRevich: In the latest PR I downgraded it to 1.20, cause in all the workflows we use 1.20.3 version, so there were failing

dabasov (approved):
lgtm

PR #1541, opened 2024-06-25
Feature: add missing attribute for WZCN burn event response entity

YarikRevich:
Related to the following issue: https://github.com/0chain/web-apps/issues/11975


dabasov (approved):
lgtm

PR #1592, opened 2024-08-30
Fix: failing dex transaction

YarikRevich:


zcncore/transaction.go
```diff
@@ -1072,13 +1072,10 @@ func (t *Transaction) Verify() error {
 
 				*t.txn = tt
 				txStatus := tt.Status
+
 				switch txStatus {
 				case 1:
-					if t.txn.TransactionType == transaction.TxnTypeSend {
-						t.completeVerifyWithConStatus(StatusSuccess, int(Success), string(output), nil)
-					} else {
-						t.completeVerifyWithConStatus(StatusSuccess, int(Success), tt.TransactionOutput, nil)
```
dabasov, line 1080: do we have an output for send transactions?

PR #1785, opened 2025-08-11
Exposed functionality to initiate and cancel reapair for mobile

### LINKED ISSUE #1786: Expose operations for repairing allocation and cancel an ongoing repair via mobilesdk
# Issue

The android build using gomobilesdk does not have RepairAllocation operation, it is present in IOS build

The resolution is to expose the RepairAllocation and CancelRepairAllocation methods that perform the actions of repair an allocation and cancel an ongoing repair operation.

# Details

The WASM implemtation used by IOS and the functionality of repairAllocation which we have to mirror in gosdk is [here](https://github.com/0chain/gosdk/blob/staging/wasmsdk/blobber.go).

wagiejack:
# Changes
- Exposed RepairAllocation and CancelRepair functions in mobilesdk to match the exposure in IOS via WASM

# Fixes
- Android apps can now initiate repair operations on storage allocations, this feature was missing compared to its IOS counterpart which used WASM to achieve the same functionality

**NOTE**:- We will not get progess updates as we are passing nil in the RepairAlloc function which is contrary to what is being done in WASM used by ios, if required in Android, we can implement that change as well later. 

## System Tests

The test for mobilesdk is successful on local as attached

image

Apart form the following legacy failures unrelated to gosdk mobile, there are no failures, the legacy failures are as follows:
 - TestAllocation_DownloadFileToFileHandler Panic - **EXISTING ISSUE**
 - 
image

The panic in TestAllocation_DownloadFileToFileHandler is not related to recent changes. This is an existing issue caused by

    - **Root Cause:** The panic occurs in GenerateOwnerSigningKey function at line 376 in zboxcore/sdk/blobber_operations.go
    - **Error:** panic: runtime error: slice bounds out of range [:32] with capacity 0
- zcncore Build Failure - **EXISTING ISSUE**
- 
image

The zcncore build failure is also not related to recent changes.

## Pre-Requisites for integration

These changes must be locally tested by the android team and post a thumbs-up only, we can merge these changes.

### DIFF
--- a/mobilesdk/sdk/sdk.go
+++ b/mobilesdk/sdk/sdk.go
@@ -30,6 +30,8 @@ import (
 
 var nonce = int64(0)
 
+var allocationIDRequired = errors.Errorf("Allocation ID is required")
+
 type Autorizer interface {
 	Auth(msg string) (string, error)
 }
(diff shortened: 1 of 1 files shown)


wagiejack:
@devpixel91 could you please let me know if it is possible to the version of gosdk from this PR locally, and if so, could you please test and share the results, Thanks!

Hitenjain14 (approved):

service-0chain:
| config  | value |
| tests skipped  | false  |
| system-tests  |   |
| gosdk  | staging  |
| zboxcli  |   |
| zwalletcli  |   |
| 0chain  |   |
| blobber  |   |
| authorizer  |   |
| 0box  |   |
| zvault  |   |
| zauth-server  |   |
| 0dns  |   |

